PR #3, opened 2023-01-08
OpenAPI REST API 0.1.0

mickel8:
Open with https://editor-next.swagger.io/


mickel8:
Response status codes are TBD

openapi/openapi.yml
```diff
@@ -0,0 +1,246 @@
...
+    url: http://www.apache.org/licenses/LICENSE-2.0.html
+  version: 0.1.0
+
+servers:
+  - url: https://localhost/api/v1
+tags:
+  - name: room
+    description: Manage rooms
+paths:
+  /room:
```
bblaszkow06, line 17: Just a thought - with Jellyfish being a high-level API maybe we should consider some other name? Especially for scenarios with an HLS broadcast of a few (or one) presenters.

openapi/openapi.yml
```diff
@@ -0,0 +1,246 @@
...
+          application/json:
+            schema:
+              $ref: '#/components/schemas/ProducerConfig'
+      responses:
+        '200':
+          description: successful operation
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Producer'
```
daniel-jodlos, line 151: These two endpoints feel like basically the same thing and a little too low-level for the high-level API that we hoped for. I think I would favor a simple `endpoint` here instead.

And I don't really see the benefit in keeping separating consumers and producers on this level. Is it permissions? Endpoints, as far as I can see, are not peers and would be a well-defined and trusted piece of code :thinking: 

I also have another problem here - this doesn't facilitate an endpoint that is both producer and consumer. Of course, we don't plan to provide it, but if someone decided to add it, this API makes it rather tricky.

openapi/openapi.yml
```diff
@@ -0,0 +1,246 @@
...
+        id:
+          type: integer
+          format: int64
+        type:
+          $ref: '#/components/schemas/ProducerType'
+    ConsumerConfig:
+      type: object
+      properties: 
+        type:
+          $ref: '#/components/schemas/ConsumerType'
```
daniel-jodlos, line 202: I think we also need a space for endpoint-specific configuration, it's a very basic feature and I think we will actually need it for HLS for example.

Also, how about allowing to set custom id? This might be a nice feature for the business logic to get a simple "hls" as an id instead of 64, or whatever it is, random letters.

openapi/openapi.yml
```diff
@@ -0,0 +1,246 @@
...
+    ProducerConfig:
+      type: object
+      properties:
+        type:
+          $ref: '#/components/schemas/ProducerType'
+    Producer:
+      type: object
+      description: "
+        Producer is a server side entity that publishes a track.
+        Example producer is RTSP, which connects to a remote camera
```
daniel-jodlos, line 188: I don't think referencing an endpoint that literally doesn't exist yet is a great idea

openapi/openapi.yml
```diff
@@ -0,0 +1,246 @@
...
+                $ref: '#/components/schemas/Producer'
+components:
+  schemas:
+    RoomConfig:
+      type: object
+      properties:
+        maxPeers:
+          type: integer
+          format: int64
+          example: 10
```
daniel-jodlos, line 160: Aren't we missing infiinty here?

daniel-jodlos (changes requested):

openapi/openapi.yml
mickel8: I thought the same but nothing better came to my mind. Thought about: cluster, chain, container

openapi/openapi.yml
mickel8: Hmm :thinking: 

> These two endpoints feel like basically the same thing and a little too low-level for the high-level API that we hoped for

I don't see any difference between having two different HTTP endpoints and one endpoint where you have to specify what kind of given engine endpoint type you want e.g. (hls_producer, hls_consumer). There are still terms `producer` `consumer` 

> I also have another problem here - this doesn't facilitate an endpoint that is both producer and consumer. Of course, we don't plan to provide it, but if someone decided to add it, this API makes it rather tricky.

The same problem is with one HTTP endpoint. How would you add an HLS engine endpoint that is both producer and consumer? It has to have some name. When we find a name for it we can create third HTTP endpoint with similar generic name

The main benefit of having separate endpoints for producers and consumers is that it is easier to distinguish from them. There is a separate enum for producers and a separate enum for consumers so it's easier to find out what producers and what consumers there are. They are also split when returning room description.

I am curious what other folks think

openapi/openapi.yml
mickel8: It's mostly for easier understanding the need for producers

openapi/openapi.yml
mickel8: Probably yes To be honest I wasn't to precies when it comes to some properites types and response status codes :roll_eyes:

openapi/openapi.yml
daniel-jodlos: > The same problem is with one HTTP endpoint. How would you add an HLS engine endpoint that is both producer and consumer? It has to have some name. When we find a name for it we can create a third HTTP endpoint with a similar generic name

1. Engine endpoints don't even have a concept of being a `producer` or `consumer`, they are what you make it - so this instantly doesn't fit the deal.
2. You will need to write basically duplicate code in the handler for these two (or three!!!) routes, or do some path parameter magic instead of writing a simple endpoint
3. Any SDK for Jellyfish will either end up with 2 functions for adding endpoints or with just one, working around that quirk.
4. It's really weird and potentially confusing. I just spoke with Paweł Kruczkiewicz about that and I found out that for him, the current HLS endpoint should count as a producer because it produces a HLS stream... So, this is not so obvious. I think this is a reason enough to use one endpoint and distinguish it with "type".

The way I see it, this very much complicates a very basic action of adding a server-side endpoint

openapi/openapi.yml
```diff
@@ -0,0 +1,246 @@
...
+    ConsumerConfig:
+      type: object
+      properties: 
+        type:
+          $ref: '#/components/schemas/ConsumerType'
+    Consumer:
+      type: object
+      description: "
+        Consumer is a server side entity that subscribes to tracks
+        published by producers or peers and process them.
```
Rados13, line 207: I think for consistency we should add examples of consumers.

Rados13 (changes requested):
Also, you forgot about endpoints regarding authorization and each endpoint will probably have an additional response code 401. You also forgot about deleting endpoints for peer, producer, and consumer.

openapi/openapi.yml
```diff
@@ -0,0 +1,246 @@
...
+              $ref: '#/components/schemas/PeerConfig'
+      responses:
+        '200':
+          description: successful operation
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Peer'
+        '500':
+          description: reached max number of peers
```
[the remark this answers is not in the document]
daniel-jodlos: Btw, isn't it the same error code that you get when you're handler raises? I'm pretty sure it is.

openapi/openapi.yml
mickel8: IMO, room is the best option at the moment

openapi/openapi.yml
```diff
@@ -0,0 +1,210 @@
...
+      enum:
+        - webrtc 
+
+    ComponentType:
+      type: string
+      enum:
+        - rtsp_producer
+        - file_producer
+        - hls_consumer
+        - file_consumer
```
mickel8, line 210: IMO we should introduce ComponentKind that will specify whether a component is a prodcuer or a consumer

openapi/openapi.yml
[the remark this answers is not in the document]
daniel-jodlos: So I don't think we should be using it here. 503 looks good for this use case, MDN literally mentions "the server is overloaded" in the description.

openapi/openapi.yml
daniel-jodlos: We already mention it in the name. What exactly do you have in mind and where do you want to use it?

daniel-jodlos:
Updated Open API specification describing existing API was merged in #20

PR #5, opened 2023-01-24
[RTC-77] Basic REST API

Rados13:


lib/jellyfish_web/views/room_view.ex
```diff
@@ -0,0 +1,23 @@
...
+
+  def render("show.json", %{room: room}) do
+    %{data: render_one(room, RoomView, "room.json")}
+  end
+
+  def render("room.json", %{room: room}) do
+    %{
+      id: room.id,
+      config: room.config,
+      components: ComponentView.render_dict(room.components),
```
LVala, line 19: Couldn't we use `ComponentView.render("index.json", %{peers: peers}`. Creating separate function to do the same thing seems unnecessary (I'm not sure if it does the same thing tho, maybe I don't understand something).

test/jellyfish_web/controllers/component_controller_test.exs
```diff
@@ -0,0 +1,72 @@
...
+    test "renders component when data is valid", %{conn: conn, room_id: room_id} do
+      conn = post(conn, Routes.component_path(conn, :create, room_id), component_type: "hls")
+      assert %{"id" => id} = json_response(conn, 201)["data"]
+
+      conn = get(conn, Routes.room_path(conn, :show, room_id))
+
+      assert %{
+               "id" => ^room_id,
+               "components" => [
+                 %{"id" => ^id, "type" => "Elixir.Membrane.RTC.Engine.Endpoint.HLS"}
```
LVala, line 20: Shouldn't the component type here be `"hls"`? I don't think we should return elixir module names as types.

LVala (changes requested):

test/jellyfish_web/controllers/component_controller_test.exs
Rados13: Nope because we want to get the type from the engine.

lib/jellyfish/component.ex
```diff
@@ -0,0 +1,71 @@
...
+  def validate_component_type(component_type) do
+    case component_type do
+      "file_reader" -> {:ok, :file_reader}
+      "hls" -> {:ok, :hls}
+      _other -> :error
+    end
+  end
+
+  @spec create_component(component_type(), any(), any()) :: t()
+  def create_component(component_type, _options, room_options) do
```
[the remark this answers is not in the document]
Rados13: I still don't know what parameters should be passed from the Room state to all endpoints.

lib/jellyfish/peer.ex
```diff
@@ -0,0 +1,97 @@
...
+  @typedoc """
+  This module contains:
+  * `id` - peer id
+  * `type` - type of peer
+  * `engine_endpoint` - engine endpoint for this peer
+  """
+  @type t :: %__MODULE__{
+          id: id,
+          type: peer_type(),
+          engine_endpoint: struct() | atom()
```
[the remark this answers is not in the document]
Rados13: Yes if the engine endpoint doesn't have any options.

lib/jellyfish/room.ex
```diff
@@ -0,0 +1,186 @@
...
+  def handle_call({:add_peer, peer_type}, _from, state) do
+    if Enum.count(state.peers) == state.config.max_peers do
+      {:reply, {:error, :reached_peers_limit}, state}
+    else
+      options = %{engine_pid: state.engine_pid, network_options: state.network_options}
+      peer = Peer.create_peer(peer_type, options)
+
+      state = put_in(state, [:peers, peer.id], peer)
+
+      :ok = Engine.add_endpoint(state.engine_pid, peer.engine_endpoint, endpoint_id: peer.id)
```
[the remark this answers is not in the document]
Rados13: [As you can see](https://github.com/jellyfish-dev/membrane_rtc_engine/blob/8b6c2276bc88c7364769f92c51121948ed27a9db/lib/membrane_rtc_engine/engine.ex#L286-L298) always return `:ok` or raise an error, so it will never return `:error`

lib/jellyfish/room.ex
```diff
@@ -0,0 +1,186 @@
...
+  end
+
+  def handle_call({:remove_component, component_id}, _from, state) do
+    {result, state} =
+      if Map.has_key?(state.components, component_id) do
+        {_elem, state} = pop_in(state, [:components, component_id])
+        {:ok, state}
+      else
+        {:error, state}
+      end
```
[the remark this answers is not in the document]
Rados13: Yup I forgot about it.

lib/jellyfish/room.ex
```diff
@@ -0,0 +1,186 @@
...
+
+    turn_cert_file =
+      case Application.fetch_env(:jellyfish, :integrated_turn_cert_pkey) do
+        {:ok, val} -> val
+        :error -> nil
+      end
+
+    turn_mock_ip = Application.fetch_env!(:jellyfish, :integrated_turn_ip)
+
+    # turn_ip = if @mix_env == :prod, do: {0, 0, 0, 0}, else: turn_mock_ip
```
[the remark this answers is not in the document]
Rados13: I am not sure about it. Because we use this line in Videoroom, but it doesn't make much sense to me, I decided to leave it until we create a prototype of jellyfish.

lib/jellyfish_web/controllers/component_controller.ex
```diff
@@ -0,0 +1,61 @@
...
+      {:error, _} ->
+        conn
+        |> put_resp_content_type("application/json")
+        |> put_status(400)
+        |> json(%{errors: "Not proper component type"})
+
+      {{:ok, _component_type}, :not_found} ->
+        conn
+        |> put_resp_content_type("application/json")
+        |> put_status(400)
```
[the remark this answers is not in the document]
Rados13: This case is for a situation when you try to add a component to not existing room. Case from the API is for a situation when you pass an unknown `component_type`.

lib/jellyfish_web/controllers/peer_controller.ex
```diff
@@ -0,0 +1,38 @@
...
+      {:error, _} ->
+        FallbackController.error_json_respond(conn, 400, "Not proper peer_type")
+
+      {{:ok, _peer_type}, :not_found} ->
+        FallbackController.error_json_respond(conn, 400, "Room not found")
+
+      {{:ok, peer_type}, room_pid} ->
+        case Room.add_peer(room_pid, peer_type) do
+          {:error, :reached_peers_limit} ->
+            FallbackController.error_json_respond(conn, 503, "Reached peers limit in room")
```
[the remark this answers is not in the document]
Rados13: You are right.

lib/jellyfish_web/controllers/room_controller.ex
```diff
@@ -0,0 +1,55 @@
...
+
+  alias Jellyfish.RoomService
+  alias Jellyfish.Room
+
+  action_fallback JellyfishWeb.FallbackController
+
+  def index(conn, _params) do
+    rooms =
+      :rooms
+      |> :ets.tab2list()
```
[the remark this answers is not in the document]
Rados13: It would be more idiomatic but it would also be a bottleneck because pretty much all requests would have to go through RoomService.

lib/jellyfish_web/views/component_view.ex
```diff
@@ -0,0 +1,25 @@
+defmodule JellyfishWeb.ComponentView do
+  use JellyfishWeb, :view
+  alias JellyfishWeb.ComponentView
+
+  def render("index.json", %{component: component}) do
+    %{data: render_many(component, ComponentView, "component.json")}
+  end
```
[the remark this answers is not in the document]
Rados13: It was autogenerated by phoenix and I think in the future it can be useful.

lib/jellyfish_web/views/room_view.ex
Rados13: I think it could be a little bit problematic that the components field is a dictionary, but I can try to change it.

lib/jellyfish/component.ex
```diff
@@ -0,0 +1,71 @@
...
+  * `component_type` - type of component
+  * `engine_endpoint` - engine endpoint for this component
+  """
+  @type t :: %__MODULE__{
+          id: id,
+          type: component_type(),
+          engine_endpoint: struct() | nil
+        }
+
+  @spec new(component_type :: atom()) :: t()
```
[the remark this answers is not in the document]
mickel8: I would favor deleting param names. I believe that's how Elixir does it

lib/jellyfish/component.ex
```diff
@@ -0,0 +1,71 @@
...
+  def new(component_type) do
+    %__MODULE__{
+      id: UUID.uuid4(),
+      type: component_type
+    }
+  end
+
+  @spec validate_component_type(component_type :: String.t()) ::
+          {:ok, component_type()} | :error
+  def validate_component_type(component_type) do
```
mickel8, line 43: This breaks SRP a little bit as not only it validates but also converts

lib/jellyfish/peer.ex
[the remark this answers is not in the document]
mickel8: We should use `Membrane.ChildrenSpec.child_definition_t` or create our type that will wrap membrane type. Can it be that engine_endpoint is `nil`?

lib/jellyfish/peer.ex
```diff
@@ -0,0 +1,97 @@
...
+  @spec new(peer_type :: peer_type()) :: t()
+  def new(peer_type) do
+    %__MODULE__{
+      id: UUID.uuid4(),
+      type: peer_type
+    }
+  end
+
+  @spec validate_peer_type(peer_type :: String.t()) :: {:ok, peer_type()} | :error
+  def validate_peer_type(peer_type) do
```
mickel8, line 40: As above

lib/jellyfish/room.ex
```diff
@@ -0,0 +1,186 @@
+defmodule Jellyfish.Room do
+  @moduledoc """
+  Module representing room.
+
```
mickel8, line 4:
```suggestion
```

lib/jellyfish/room_service.ex
```diff
@@ -0,0 +1,72 @@
...
+    GenServer.start(__MODULE__, init_arg, opts)
+  end
+
+  def start_link(opts) do
+    GenServer.start_link(__MODULE__, [], opts)
+  end
+
+  @impl true
+  def init(_) do
+    :ets.new(:rooms, [:protected, :set, :named_table])
```
mickel8, line 15: Why ETS?

lib/jellyfish/room_service.ex
```diff
@@ -0,0 +1,72 @@
...
+  @impl true
+  def handle_call({:delete_room, _room_id}, _from, state) do
+    {:reply, :not_found, state}
+  end
+
+  @spec find_room(room_id :: String.t()) :: pid() | :not_found
+  def find_room(room_id) do
+    case :ets.lookup(:rooms, room_id) do
+      [{_room_id, room_pid} | _] -> room_pid
+      _not_found -> :not_found
```
mickel8, line 55: Follow  `Enum.find` and return `nil`

lib/jellyfish/room_service.ex
```diff
@@ -0,0 +1,72 @@
...
+    {:reply, :ok, state}
+  end
+
+  @impl true
+  def handle_call({:delete_room, _room_id}, _from, state) do
+    {:reply, :not_found, state}
+  end
+
+  @spec find_room(room_id :: String.t()) :: pid() | :not_found
+  def find_room(room_id) do
```
mickel8, line 52: Could we put public API at the top of the module?

lib/jellyfish_web/views/component_view.ex
```diff
@@ -0,0 +1,25 @@
+defmodule JellyfishWeb.ComponentView do
```
mickel8, line 1: After reading https://hexdocs.pm/phoenix/views.html#rendering-json, I don't see how we can benefit from using Views for returning JSON. In general, we don't want to render data, we just want to return JSON to the user that will execute its requests via curl, postman etc. 

Instead, I would favor implementing Jason.Econder for needed structs https://github.com/michalmuskala/jason#encoders

Using views for returning JSON seems to me like an overkill, results in a lot of boilerplate, and requires from us manual changes when Component struct changes (e.g. when adding fields)

We can ask someone else, someone who is more experienced in Phoenix

mickel8 (changes requested):
TBC

lib/jellyfish/component.ex
Rados13: I think this is more bad naming and this function should be `parse_component_type`, also function can do multiple things and is not against SRP e.g: `getAndIncrement` in `AtomicInteger` in Java.

lib/jellyfish/room.ex
LVala: I'm not sure what does that mean, to remove the `moduledoc` (well, it doesn't convey much information)?

lib/jellyfish/room_service.ex
LVala: Now it is consistent with other functions and returns tuple.

lib/jellyfish/room.ex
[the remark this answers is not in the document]
LVala: @mickel8 maybe (surely, right?) you know something about this?

LVala:
I left stuff about moving config to file to another PR as we mentioned.
Also, please look at the comments I left above.
Code coverage shows that some areas are not properly tested and I'm wondering if I should write more tests here of leave it as it is?

mix.exs
```diff
@@ -54,6 +54,21 @@ defmodule Jellyfish.MixProject do
       {:telemetry_poller, "~> 1.0"},
       {:jason, "~> 1.2"},
       {:plug_cowboy, "~> 2.5"},
+      {:uuid, "~> 1.1"},
+
+      # Membrane deps
+      {:membrane_rtc_engine,
+       github: "jellyfish-dev/membrane_rtc_engine", branch: "get_endpoints"},
```
mickel8, line 61: Could we solve this before merging?

lib/jellyfish/room.ex
mickel8: I just meant to remove that one line

lib/jellyfish/room_service.ex
mickel8: I think that the question here is whether RoomService might be overhelmed by HTTP requests. I would keep ets for now and observe how it behaves

mickel8 (changes requested):
I can see some potential for abstraction for configuring/creating RTC Engine endpoints. What do you think about following bumblebee and creating a behavior that will have just one callback called `config`.

More or less something like this?

```elixir
defmodule X do
  @callback config(opts) :: Membrane.ParentSpec.child_spec_t()
end

defmodule Component.WebRTC do
  @behavior X

  @impl true
  def config(opts), do: %WebRTC{}
end

defmodule Component do

  def new_from_string(type, opts) when type in ["hls"] do
    type = to_module(type)
    %__MODULE__{
      id: new_uuid()
      type: type,
      engine_endpoint: type.config(opts)
  }
end 
```

You can read more in bumblebee. 

https://github.com/elixir-nx/bumblebee/blob/v0.1.0/lib/bumblebee.ex#L198
https://github.com/elixir-nx/bumblebee/blob/3b58b8040b675048ab19ae1bf48c1170126b291c/lib/bumblebee/configurable.ex#L1

I am not saying we have to do this like that way. just curious your opinion

lib/jellyfish/room_service.ex
LVala: Ok, I thought that the `RoomService` process used the `:ets`, but it seems that it sort of replaces `GenServer.call/3`, which makes a lot more sense.

LVala:
The only thing left is to get rid of `membrane_rtc_engine` branch dependency and move config to file(another PR). Also, last commit solves issue with dialyzer/compiler angry about using `Mix.env()` (docker/turn ip stuff), but it probably [should be refactored](https://elixirforum.com/t/a-universal-way-to-detect-environment-in-phoenix/39934/4).

PR #6, opened 2023-01-26
[RTC-78] Setup CI

LVala:


lib/jellyfish_web.ex
```diff
@@ -50,6 +57,9 @@ defmodule JellyfishWeb do
     end
   end
 
+  @spec channel ::
+          {:use, [{:context, JellyfishWeb} | {:import, Kernel}, ...],
+           [{:__aliases__, [...], [...]}, ...]}
```
Rados13, line 62: I am not sure if these specs are useful in any way.

test/jellyfish_test.exs
```diff
@@ -1,8 +1,4 @@
 defmodule JellyfishTest do
   use ExUnit.Case
   doctest Jellyfish
-
-  test "greets the world" do
-    assert Jellyfish.hello() == :world
-  end
 end
```
Rados13, line 4: I think that currently, we could remove this file.

mix.exs
```diff
@@ -16,16 +16,25 @@ defmodule Jellyfish.MixProject do
...
-            docs: docs()
+      # hex
+      description: "Jellyfish media server",
+      package: package(),
+
+      # docs
+      name: "Jellyfish media server",
+      source_url: @github_url,
+      homepage_url: "https://membrane.stream",
+      docs: docs(),
```
mickel8, line 27: It is very unlikely that we will use `ex_doc` for documentation, could we remove it?

.credo.exs
```diff
@@ -31,7 +31,13 @@
           "apps/*/test/",
           "apps/*/web/"
         ],
-        excluded: [~r"/_build/", ~r"/deps/", ~r"/node_modules/"]
+        excluded: [
+          ~r"/_build/",
+          ~r"/deps/",
+          ~r"/node_modules/",
+          "lib/jellyfish_web.ex",
+          "lib/jellyfish_web/telemetry.ex"
```
mickel8, line 39: If the reason of ignoring `jellyfish_web.ex` and `telemetry.ex` is the requirement for specs for public functions, I would favor disabling ` Credo.Check.Readability.Specs` instead

mickel8 (changes requested):

Rados13 (approved):

lib/jellyfish_web/views/error_helpers.ex
```diff
@@ -3,14 +3,15 @@ defmodule JellyfishWeb.ErrorHelpers do
   Conveniences for translating and building error messages.
   """
 
+  @spec translate_error({any, any}) :: any
   @doc """
   Translates an error message.
   """
   def translate_error({msg, opts}) do
```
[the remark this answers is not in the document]
LVala: I also removed `error_view.ex`. We probably won't be using templates anyway.

mix.exs
```diff
@@ -16,16 +15,24 @@ defmodule Jellyfish.MixProject do
...
-            homepage_url: "https://membraneframework.org",
-            docs: docs()
+      # hex
+      description: "Jellyfish media server",
+      package: package(),
+
+      # docs
+      name: "Jellyfish media server",
+      source_url: @github_url,
+      homepage_url: "https://membrane.stream",
```
mickel8, line 25: Sounds like it can be remove too, can't it?

mickel8 (approved):

PR #7, opened 2023-01-31
Rest api improvements

LVala:


mickel8:
Actually, it's pretty hard to review those changes when I haven't reviewed the base branch yet. I would propose one of the following:
1. merge this to the base branch (open by @Rados13) and I will review the whole PR at once
2. ask @Rados13 for CR, merge this to its branch and I will reivew the whole PR at once

lib/jellyfish_web/controllers/component_controller.ex
```diff
@@ -7,54 +7,37 @@ defmodule JellyfishWeb.ComponentController do
...
-
-        conn
-        |> put_status(:created)
-        |> render("show.json", component: component)
+  def create(conn, %{"room_id" => room_id} = params) do
+    with {:ok, component_type_string} <- Map.fetch(params, "type"),
+         component_options <- Map.get(params, "options"),
+         {:ok, component_type} <- Component.validate_component_type(component_type_string) do
+      case RoomService.find_room(room_id) do
+        :not_found ->
```
Rados13, line 15: As I see it now, I think that `RoomService.find_room` should return `room_pid | {:error, :room_not_found}`.

test/jellyfish_web/controllers/peer_controller_test.exs
```diff
@@ -1,43 +1,94 @@
...
+
+    test "deletes not existing peer", %{conn: conn, room_id: room_id} do
+      peer_id = "test123"
+      conn = delete(conn, Routes.peer_path(conn, :delete, room_id, peer_id))
+
+      assert json_response(conn, :not_found)["errors"] ==
+               "Peer with id #{peer_id} doesn't exist"
+    end
+
+    test "deletes component from not exisiting room", %{conn: conn, peer_id: peer_id} do
```
Rados13, line 80:
```suggestion
    test "deletes peer from not exisiting room", %{conn: conn, peer_id: peer_id} do
```

test/jellyfish_web/controllers/room_controller_test.exs
```diff
@@ -8,42 +8,44 @@ defmodule JellyfishWeb.RoomControllerTest do
...
-      assert %{"id" => id} = json_response(conn, 201)["data"]
+      conn = post(conn, Routes.room_path(conn, :create), maxPeers: 10)
+      assert %{"id" => id} = json_response(conn, :created)["data"]
 
       conn = get(conn, Routes.room_path(conn, :show, id))
 
       assert %{
                "id" => ^id,
-               "config" => %{"max_peers" => 10},
+               "config" => %{"maxPeers" => 10},
```
Rados13, line 24: Why do you use camel case?

LVala: Doesn't JSON use camelCase by convention? Also, that how it was specified [here](https://github.com/jellyfish-dev/jellyfish/pull/3/files#diff-f26912d4d2c7811bd456bf4f854b781a6a26384ea258dc55dfb592953cc4f9faR141).

Rados13: Yeah you are right.

lib/jellyfish/room.ex
```diff
@@ -57,23 +56,27 @@ defmodule Jellyfish.Room do
 
   @impl true
   def handle_call(:state, _from, state) do
-    engine_endpoints = Engine.get_endpoints(state.engine_pid)
+    active_endpoints =
+      state.engine_pid
+      |> Engine.get_endpoints()
+      |> Enum.map(& &1.id)
+      |> MapSet.new()
```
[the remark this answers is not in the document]
LVala: Well, I changed that the returned type of peer/component is the Jellyfish type, not Elixir module name, so I only need the list of actually active endpoints from RTC Engine in order to check, if peers/components in Jellyfish state exist (so in theory `MapSet` is better to check if it contains some value, but it's negligible, I can change it back to list if you prefer).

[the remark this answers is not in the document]
Rados13: It's ok.

LVala:
Im not sure whether we can get rid of `jellyfish_web/views/error_view.ex` as it seems that we don't need it. `error_helpers.ex` is used by the `use JellyfishWeb, :view`, do it has to stay.

Rados13 (approved):

PR #8, opened 2023-02-14
[RTC-90] Signalling architecture

LVala:
Looking for opinions and suggestions on signalling architecture. Feel free to comment. Nice GitHub view [here](https://github.com/jellyfish-dev/jellyfish/blob/signaling-arch/docs/signaling_architecture.md).


LVala:
Added document about authentication architecture to this PR, as the topics are intertwined quite a bit, so there's some overlap in the documents.
Authentication [here](https://github.com/jellyfish-dev/jellyfish/blob/signaling-arch/docs/auth-arch.md)

docs/auth-arch.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Using **BE** as a middleman
+
+The same assumptions and first 3 steps as in the previous example:
+
+4) **BE** responds to **CL** that it was authorized and that it can send signaling messages to **BE**.
+
+Problems:
+
+- in current implementation of `membrane_rtc_engine`, endpoint should be created after signaling connection was established. In previous example it was obvious:
```
daniel-jodlos, line 46: > in current implementation of `membrane_rtc_engine`, endpoint should be created after signaling connection was established
Yes, it would probably be a good idea, but preamptively creating an endpoint would be acceptable if you made sure to clean it up - the endpoint awaits `join` media event before becoming fully active.

docs/signaling_architecture.md
```diff
@@ -0,0 +1,97 @@
...
+- tag every signaling message (Media Event) with e.g. `peer_id`, requires only one WebSocket connection between **BE** and **JMS** (should SDK do this, or should it be the users responsibility?).
+
+**Advantages:**
+
+- Easy to implement (from our, Jellyfish developers, perspective).
+- **CL** does not require additional connection and authentication, everything is handled by **BE**.
+
+**Disadvantages:**
+
+- Harder to implement by the user, much more error prone (isn't that the point of Jellyfish to make it as simple as possible?).
```
daniel-jodlos, line 71: I think we were also aiming to minimize boilerplate, and in this approach, you would basically have to implement the signaling channel yourself, which generates boilerplate code.

mickel8 (commented):
When it comes to signalling architecture I would favor direct connection between CL and JF. Mostly becasue:
* users might incorrectly implement signalling and our ability to help them might be limited when their code is proprietary 
* it is easier and faster to setup anything that works with direct connection
* we will have to duplicate signaling implementation in each product
* I don't see so many RC that we will have to solve. Anyway, I think it's easier to fix RC than to explain everyone how to implement signalling. Mostly because RC can be noticed by users on their own and won't require our help
* direct connection seems to me to be a better choice in the context of SaaS. Actually, I am not sure wheter SaaS architecture is achievable when using BE as a middleman for signalling :thinking:

docs/signaling_architecture.md
```diff
@@ -0,0 +1,97 @@
...
+3) **JMS** responds positively.
+4) **BE** responds to client with generated token.
+5) **CL** opens WebSocket connection to **JMS**, flow of signaling messages begins.
+
+**JMS** can diferentiate between clients by the opened WebSocket connection so it knows who is the sender of
+incoming Media Events and where to send generated Media Events.
+
+**Advantages:**
+
+- Easier option, the user doesn't have to do much more than passing some secret (authentication token) to the **CL**, client SDK handles the rest.
```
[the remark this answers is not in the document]
LVala: Right, I wrote that before getting into authentication and misunderstood the topic.

daniel-jodlos:
>  I don't see so many RC that we will have to solve

Neither do I, but I would be willing to accept an approach where we allow replacing the signalling with your custom implementation, if we get an example of a Race Condition that cannot easily be fixed in the Direct Connection.

PR #10, opened 2023-02-20
[RTC-82] Authentication

LVala:
Moved document about authentication from the other PR here. Nice GitHub markdown view [here](https://github.com/jellyfish-dev/jellyfish/blob/auth-arch/docs/authentication_architecture.md).

### DIFF
--- a/docs/authentication_architecture.md
+++ b/docs/authentication_architecture.md
@@ -74,5 +74,5 @@ No matter who generates the tokens, effort from the user comes to passing the to
 
 We will be using **JF** to generate tokens, as it makes it easier to maintain (we don't need to implement token generation in all of our SDKs) and we keep all of the logic responsible for generation and validation together.
 Also, **BE** authentication might not be very difficult: they may just share a common secret and use it directly to authenticate (e.g. via HTTP authorization request header), especially when **JF** and **BE** are in the same internal
-network, which will be a common case. Using **BE** to generate tokens could also complicate things when using more than one instance of **JF** and make it harder to scale and make modifications related to tokens (need to redeploy
+network, which will be a common case. Using **BE** to generate tokens could also make it harder to scale and make modifications related to tokens (need to redeploy
 both **JF** and **BE** in such case).
--- a/docs/authentication_architecture.md
+++ b/docs/authentication_architecture.md
@@ -74,5 +74,4 @@ No matter who generates the tokens, effort from the user comes to passing the to
 
 We will be using **JF** to generate tokens, as it makes it easier to maintain (we don't need to implement token generation in all of our SDKs) and we keep all of the logic responsible for generation and validation together.
 Also, **BE** authentication might not be very difficult: they may just share a common secret and use it directly to authenticate (e.g. via HTTP authorization request header), especially when **JF** and **BE** are in the same internal
-network, which will be a common case. Using **BE** to generate tokens could also make it harder to scale and make modifications related to tokens (need to redeploy
-both **JF** and **BE** in such case).
+network, which will be a common case. Using **BE** to generate tokens could also make it harder make modifications related to tokens (need to redeploy both **JF** and **BE** in such case).
--- a/docs/authentication_architecture.md
+++ b/docs/authentication_architecture.md
@@ -74,4 +74,4 @@ No matter who generates the tokens, effort from the user comes to passing the to
 
 We will be using **JF** to generate tokens, as it makes it easier to maintain (we don't need to implement token generation in all of our SDKs) and we keep all of the logic responsible for generation and validation together.
 Also, **BE** authentication might not be very difficult: they may just share a common secret and use it directly to authenticate (e.g. via HTTP authorization request header), especially when **JF** and **BE** are in the same internal
-network, which will be a common case. Using **BE** to generate tokens could also make it harder make modifications related to tokens (need to redeploy both **JF** and **BE** in such case).
+network, which will be a common case. Using **BE** to generate tokens could also make it harder to make modifications related to tokens (need to redeploy both **JF** and **BE** in such case).
--- a/docs/authentication_architecture.md
+++ b/docs/authentication_architecture.md
@@ -37,6 +37,8 @@ Problems:
 
 ### Using **BE** as a middleman
 
+**Note: direct signaling option have been chosen so this paragraph is not applicable at the moment.**
+
 The same assumptions and first 3 steps as in the previous example:
 
 4) **BE** responds to **CL** that it was authorized and that it can send signaling messages to **BE**.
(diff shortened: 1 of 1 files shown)


docs/authentication_architecture.md
```diff
@@ -0,0 +1,71 @@
...
+This approach only makes sense when using direct signaling, otherwise tokens are not necessary at all (except for **BE** - **JF** connection).
+
+### **JF**
+
+In this approach **JF** creates the tokens (when using `add_peer`, token is send in the response), so user only needs to pass it to the **CL**.
+You also might need to pass expected token presmissions to **JF**.
+Despite that, we still need a way to authenticate **BE** - **JF** connection. Possible solutions:
+
+- create JWT on **BE** side anyway (in such case you might as well use the first approach in order not to split logic responsible for token generation between Server SDK and Jellyfish),
+- create JWT (or some other token type) once and use it in configuration (makes it easier to change **BE** permissions, if that's ever necessary, but the token never expires, I'm not sure whether that's a problem, also **BE** doesn't need to know the secret).
```
daniel-jodlos, line 69: We probably don't want to implement that in the 0.1.0, but we should think about secret rotation. Some security certifications will require the secrets to be rotated once a few months or in case of a security breach.

Also, I think I would prefer an option where you can limit the permissions of the BE. For once, applications should always have a set of permissions that gives them access only to the features that they actually use. But even more importantly, if we implemented secret rotation, the permissions for this should only be given to the process that performs the actual rotation.  And if the BE has a token with limited permissions, it cannot know the secret used to sign that token - if it did, it could generate a new token with higher permissions and then we might as well not bother about limiting the permissions.

mickel8: > And if the BE has a token with limited permissions, it cannot know the secret used to sign that token - if it did, it could generate a new token with higher permissions and then we might as well not bother with limiting the permissions.

Any ideas how to achieve this?

To me it sounds like we don't trust our own backend and would like to have some other security entity over it

mickel8 (commented):
Phoenix has built-in Token module which could be used for generating and veryfing tokens if we decided to the second apprach (the one where server generates token)

https://hexdocs.pm/phoenix/Phoenix.Token.html#encrypt/4

Using this module we could avoid implementing token generation in each server SDK (elixir, go, rust, java, etc.) and just rely on Phoenix.

My only concern is that I don't know how to authenticate BE in such a case. I assume that both BE and JF would have to be configured with some secret but what's next?

mickel8:
> My only concern is that I don't know how to authenticate BE in such a case. I assume that both BE and JF would have to be configured with some secret but what's next?

We of course pass it in HTTP request headers :man_facepalming: 

Looking at livekit, they have chosen option where tokens are created by SDK and they create a new token every time they want to modify room state https://github.com/livekit/server-sdk-go/blob/main/roomclient.go.

So for every single operation like createRoom, removeRoom etc. they create a new token with different permissions.

I am not sure whether we need such a security at the beginning and the solution where server generates tokens seems to be easier to implement as we don't have to duplicate token generation in every server SDK so maybe we should start with simpler solution, discover its limits and then move to the more flexible one?

bblaszkow06:
I'm also in favour of generating tokens by Jellyfish. The most important thing for me is that it ensures the same service generates and verifies the tokens without sharing a secret.
Otherwise, we would be forced to keep them in sync and I imagine it could be a mess in a few scenarios:
* More than one JF - we'd want the token to be working for only one instance
* Auto-scaling JF - we'd either need one token for all JF instances or update each backend to have a new secret for added JF
* Changing anything related to tokens (either the secret or the way it is generated) would require redeployment of both JF and backends

This approach won't prevent us from generating tokens with very granular permissions

For BE <> JF auth a generated token passed by config seems enough for a start, as when deployed I'd expect access to the SDK endpoint to be limited to the internal network or even local (e.g. when running as 2 Docker containers on the same machine we'd only expose backend's endpoint and client endpoint + media of JF server)

bblaszkow06:
BTW we might in future need a different auth options for different scenarios, Janus has 3: https://janus.conf.meetecho.com/docs/auth.html

mickel8:
>  The most important thing for me is that it ensures the same service generates and verifies the tokens without sharing a secret.

@bblaszkow06  The initial idea was that we configure JF and BE with the same secret and this secret is used for authorizing HTTP requests (via authorization request header). But if I understand you correctly, you would like to configure BE with pre-generated JF token.  Who would be responsible for generating this token in such a case? Would someone have to run JF just to call some function to obtain the token? 

>  as when deployed I'd expect access to the SDK endpoint to be limited to the internal network or even local (e.g. when running as 2 Docker containers on the same machine we'd only expose backend's endpoint and client endpoint + media of JF server)

While deploying on the same machine should be possible, I belive it shouldn't be our target/main use-case as we want to keep multimedia server as performant as possible. The scenario with private network is a good point but I think we should be prepared that JF and BE might be deployed in completely different networks. I can imagine that some people will use machines from different providers. 

> BTW we might in future need different auth options for different scenarios, Janus has 3: https://janus.conf.meetecho.com/docs/auth.html

I would really like to avoid this :pray:  At the moment, it sounds to me like Janus has really bad API and that's the reason of having 3 different authentication methods but I might not be aware of some things yet.

mickel8:
Another question we should answer - who will be allowed to open WS connection for receiving real-time notifications from JF like "a new peer joined", "recording has started", etc.? Only the BE that was configured wit hsome secret/token or we would like to add a new HTTP endpoint for generating token with permissions to open WS connection for observability.

bblaszkow06:
> But if I understand you correctly, you would like to configure BE with a pre-generated JF token. Who would be responsible for generating this token in such a case? Would someone have to run JF just to call some function to obtain the token?

No, that's not what I meant - for BE <> JF auth it doesn't make sense to do this. The secret can be the token, you can generate it by punching the keyboard with your fist as long as both JF & BE are configured with it.

However, for client JF tokens it would be best to keep that logic (generation + validation) within one service and these were the tokens I was referring to in the first sentence.

> While deploying on the same machine should be possible, I believe it shouldn't be our target/main use case as we want to keep multimedia server as performant as possible.

Well, the amount of work done by the backend is negligible and it might be easier to scale whole stacks (BE + JF) than having n to m infrastructure - it complicates things a lot.

> The scenario with private network is a good point but I think we should be prepared that JF and BE might be deployed in completely different networks. I can imagine that some people will use machines from different providers.

I'd compare this to having a DB - the authentication is based on username/password (You can consider our BE<>JF secret token to be "username:password") and it is a common thing to have additional security for it, e.g. hiding it inside private network or setting up firewall to whitelist only specific IP addresses.
Besides, even for different providers setting up a private network is possible (via tunnelling)

Anyway, my point is we don't need a sophisticated auth for the backend without a sensible use case.

mickel8:
I think that we have consensus here and can try to merge this PR. @LVala in a free time, could you please write a short summary explaining which solution we've picked and why?

docs/authentication_architecture.md
```diff
@@ -0,0 +1,77 @@
...
+
+No matter who generates the tokens, effort from the user comes to passing the token to **CL** (in direct signaling, otherwise no need to do anything except for creating the signalling connection).
+
+## Consclusion
+
+We will be using **JF** to generate tokens, as it makes it easier to maintain (we don't need to implement token generation in all of our SDKs) and we keep all of the logic responsible for generation and validation together.
+Also, **BE** authentication might not be very difficult: they may just share a common secret and use it directly to authenticate, especially when **JF** and **BE** are in the same internal network, which will be a common case.
+Using **BE** to generate tokens could also complicate things when using more than one instance of **JF** and make it harder to scale and make modifications related to tokens (need to redeploy both **JF** and **BE** in such case).
```
mickel8, line 77: Could we elaborate this? IMO, the only difference is that we either use bare secret for authentication or create some token using this secret. In both cases we need some secret so I am not sure how the second approach (when generating token on BE) affects scenario with multiple JF

mickel8 (changes requested):

bblaszkow06 (approved):

mickel8 (approved):

PR #9, opened 2023-02-15
[RTC-79] Setup docker CD

LVala:

PR #11, opened 2023-02-23
RTC-85 Deploy to fly.io

roznawsk:
The current jellyfish version has been deployed to fly.io however, it is unlikely that it will work once we create an actual application which is streaming media. 

Some people stumbled upon problems when deploying to fly.io, and we will probably have to face them as well: https://github.com/jellyfish-dev/membrane_rtc_engine/issues/171.

Also, the Docker image used is different from the one created in [this PR](https://github.com/jellyfish-dev/jellyfish/pull/9) as it based on Ubuntu not Alpine. Ubuntu is recommended by Fly.io for reasons related to DNS.

We should wait with any work on the deployment until we have a functional multimedia application.


config/runtime.exs
```diff
@@ -66,6 +86,9 @@ config :jellyfish,
   webrtc_used: String.downcase(System.get_env("WEBRTC_USED", "true")) not in ["false", "f", "0"],
   integrated_turn_ip:
     System.get_env("INTEGRATED_TURN_IP", "127.0.0.1") |> ConfigParser.parse_integrated_turn_ip(),
+  integrated_turn_prod_ip:
+    System.get_env("INTEGRATED_TURN_PROD_ADDR", "0.0.0.0")
```
mickel8, line 90: Let's call this `INTEGRATED_TURN_LISTEN_IP`

config/runtime.exs
```diff
@@ -19,6 +19,26 @@ defmodule ConfigParser do
     end
   end
 
+  def parse_integrated_turn_prod_address(addr) do
```
mickel8, line 22: Rename to `parse_integrated_turn_ip` and remove the old function

config/runtime.exs
```diff
@@ -107,9 +130,11 @@ if config_env() == :prod do
       # Set it to  {0, 0, 0, 0, 0, 0, 0, 1} for local network only access.
       # See the documentation on https://hexdocs.pm/plug_cowboy/Plug.Cowboy.html
       # for details about using IPv6 vs IPv4 and loopback vs public addresses.
-      ip: {0, 0, 0, 0, 0, 0, 0, 0},
+      # ip: {0, 0, 0, 0, 0, 0, 0, 0},
+      ip: {0, 0, 0, 0},
```
mickel8, line 134: I don't think we should change this :thinking:

config/runtime.exs
```diff
@@ -107,9 +130,11 @@ if config_env() == :prod do
...
       port: port
     ],
+    check_origin: false,
```
mickel8, line 137: That's suspicious

mix.exs
```diff
@@ -76,7 +77,10 @@ defmodule Jellyfish.MixProject do
       {:credo, ">= 0.0.0", only: :dev, runtime: false},
 
       # Test deps
-      {:excoveralls, "~> 0.15.0", only: :test, runtime: false}
+      {:excoveralls, "~> 0.15.0", only: :test, runtime: false},
+
+      # OTEL
+      {:opentelemetry_exporter, "~> 1.0.0"}
```
mickel8, line 83: Shouldn't be needed?

mix.exs
```diff
@@ -56,7 +56,7 @@ defmodule Jellyfish.MixProject do
       {:ymlr, "~> 3.0"},
 
       # Membrane deps
-      {:membrane_rtc_engine, github: "jellyfish-dev/membrane_rtc_engine"},
+      {:membrane_rtc_engine, github: "jellyfish-dev/membrane_rtc_engine", branch: "master"},
```
mickel8, line 59: Master is the default branch

lib/jellyfish/room.ex
```diff
@@ -300,7 +300,12 @@ defmodule Jellyfish.Room do
 
         turn_mock_ip = Application.fetch_env!(:jellyfish, :integrated_turn_ip)
 
-        turn_ip = if @is_prod, do: {0, 0, 0, 0}, else: turn_mock_ip
+        turn_ip =
+          if @is_prod do
+            Application.fetch_env!(:jellyfish, :integrated_turn_prod_ip)
+          else
+            turn_mock_ip
+          end
```
mickel8, line 308:
```suggestion
        turn_ip = Application.fetch_env!(:jellyfish, :integrated_turn_listen_ip)
 ```

File: fly.toml.sample [hunk 8]
```diff
@@ -0,0 +1,46 @@
...
+    port = 443
+
+  [[services.tcp_checks]]
+    grace_period = "1s"
+    interval = "15s"
+    restart_limit = 0
+    timeout = "2s"
+
+[[services]]
+  internal_port = 50000
```
mickel8, line 42: Is that needed?

fly.toml.sample
```diff
@@ -0,0 +1,46 @@
+kill_signal = "SIGTERM"
+kill_timeout = 5
+processes = []
+
+[env]
+  INTEGRATED_TCP_TURN_PORT = "49999"
```
mickel8, line 6: I believe this port needs to be in `services.ports`, doesn't it?

fly.toml.sample
```diff
@@ -0,0 +1,46 @@
...
+  INTEGRATED_TURN_IP = "127.0.0.1"
```
mickel8, line 7: We are missing INTEGRATED_TURN_PROD_IP, and instead of  `127.0.0.1`  let's write `<ip obtained with fly ips allocate-v4>`

mickel8 (changes requested):

fly.toml.sample
```diff
@@ -0,0 +1,46 @@
...
+    interval = "15s"
+    restart_limit = 0
+    timeout = "2s"
+
+[[services]]
+  internal_port = 50000
+  protocol = "udp"
+
+  [[services.ports]]
+    port = 50000
```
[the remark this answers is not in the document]
roznawsk: But that would map all these ports to a single 50000 internal port. We cannot define (and don't have to) multiple UDP ports. They should all be forwarded, theoretiacally. In practice I noticed that any UDP port mapping must be defined in the `services` section of `fly.toml` in order for UDP to work.

fly.toml.sample
```diff
@@ -0,0 +1,46 @@
...
+
+[env]
+  INTEGRATED_TCP_TURN_PORT = "49999"
+  INTEGRATED_TURN_IP = "127.0.0.1"
+  INTEGRATED_TURN_PORT_RANGE = "50000-59999"
+  PORT = "4000"
+  WEBRTC_USED = "true"
+
+[experimental]
+  auto_rollback = true
```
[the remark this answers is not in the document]
roznawsk: Good question. The docs don't mention the purpose of this setting. This was autogenerated btw. so I think we might leave here. I can only presume that it allows for automatically reverting to a previous deployment of an app, if the current one fails.

fly.toml.sample
```diff
@@ -0,0 +1,46 @@
...
+
+[[services]]
+  http_checks = []
+  internal_port = 4000
+  processes = ["app"]
```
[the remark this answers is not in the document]
roznawsk: Not really, I just left it here, as it was autogenerated.

fly.toml.sample
roznawsk: You're right, it should be. However this variable isn't used anywhere (other than parsing it), so I am not sure if we should set it at all.

roznawsk:
> I don't see any changes in Dockerfile. Should I? Could you link an article wher Fly.io suggests using Ubuntu instead of Alpine?

`fly launch` appends these lines:
```
# Appended by flyctl
ENV ECTO_IPV6 true
ENV ERL_AFLAGS "-proto_dist inet6_tcp"
```

When generating new app with `fly launch`, without existing Dockerfile, this is the comment at the top of it:
```
# Find eligible builder and runner images on Docker Hub. We use Ubuntu/Debian
# instead of Alpine to avoid DNS resolution issues in production.
#
# https://hub.docker.com/r/hexpm/elixir/tags?page=1&name=ubuntu
# https://hub.docker.com/_/ubuntu?tab=tags
#
# This file is based on these images:
#
#   - https://hub.docker.com/r/hexpm/elixir/tags - for the build image
#   - https://hub.docker.com/_/debian?tab=tags&page=1&name=bullseye-20221004-slim - for the release image
#   - https://pkgs.org/ - resource for finding needed packages
#   - Ex: hexpm/elixir:1.14.3-erlang-25.2-debian-bullseye-20221004-slim
#
```

I didn't find this information anywhere else, though.

config/runtime.exs
```diff
@@ -59,12 +67,12 @@ config :jellyfish,
...
     System.get_env("INTEGRATED_TURN_IP", "127.0.0.1") |> ConfigParser.parse_integrated_turn_ip(),
+  integrated_turn_listen_ip:
+    System.get_env("INTEGRATED_TURN_LISTEN_IP", "0.0.0.0")
+    |> ConfigParser.parse_integrated_turn_ip(),
   integrated_turn_port_range:
     System.get_env("INTEGRATED_TURN_PORT_RANGE", "50000-59999")
     |> ConfigParser.parse_integrated_turn_port_range(),
-  integrated_tcp_turn_port:
-    System.get_env("INTEGRATED_TCP_TURN_PORT")
-    |> ConfigParser.parse_port_number("INTEGRATED_TCP_TURN_PORT"),
```
mickel8, line 67: We need to keep this and pass to the integrated turn so it takes effect. If not in this PR then in another

mickel8 (changes requested):

PR #14, opened 2023-03-01
[RTC-150/1] Server-side signalling

LVala:
This PR contains custom socket implementation and basic media Event exchange between client and server. There are some TODOs left and it's untested, but besides that it can be reviewed.


mickel8 (changes requested):
Looks really nice! I added a couple of comments but nothing really serious

lib/jellyfish_web/socket.ex
```diff
@@ -0,0 +1,75 @@
...
+
+  @impl true
+  def connect(state) do
+    with {:ok, peer_id} <- Map.fetch(state.params, "peer_id"),
+         {:ok, room_id} <- Map.fetch(state.params, "room_id"),
+         {:ok, room_pid} <- RoomService.find_room(room_id),
+         {:ok, :disconnected} <- Room.get_peer_connection_status(room_pid, peer_id) do
+      state =
+        state
+        |> Map.put(:room_pid, room_pid)
```
mickel8, line 21: If a room is registered, it should be posssible to use its id instead of pid. This way we are resiliant for room restarts - pid changes in such a case

mickel8:
@bblaszkow06 do you want to take a look too?

lib/jellyfish/room.ex
```diff
@@ -157,6 +207,62 @@ defmodule Jellyfish.Room do
...
+        )
+    end
+
+    {:noreply, state}
+  end
+
+  @impl true
+  def handle_info(%Message.EndpointCrashed{endpoint_id: endpoint_id}, state) do
+    Logger.error(
+      "RTC Engine endpoint associated with peer #{inspect(endpoint_id)} crashed, room: #{inspect(state.id)}"
```
daniel-jodlos, line 233: This would be misleading. Endpoint doesn't always represent a peer (in RTC Engine)

lib/jellyfish_web/socket.ex
```diff
@@ -0,0 +1,138 @@
...
+    case Jason.decode(encoded_message) do
+      {:ok, %{"type" => "mediaEvent", "data" => data}} ->
+        case RoomService.find_room(state.room_id) do
+          {:ok, room_pid} ->
+            send(room_pid, {:media_event, state.peer_id, data})
+
+          {:error, :room_not_found} ->
+            Logger.warn(
+              "Trying to send Media Event to room #{inspect(state.room_id)}) that does not exists "
+            )
```
daniel-jodlos, line 94: I don't think this should be possible. Both `init` and `connect` will return an error if the room doesn't exist, so you can just save the pid somewhere

lib/jellyfish/room_service.ex
```diff
@@ -41,9 +40,11 @@ defmodule Jellyfish.RoomService do
   @impl true
   def handle_call({:create_room, max_peers}, _from, state)
       when is_nil(max_peers) or is_number(max_peers) do
-    {:ok, room_pid} = GenServer.start_link(Room, max_peers)
+    {:ok, room_pid} = Room.start_link(max_peers)
```
daniel-jodlos, line 43: If I understand correctly, if the room dies, the entire room service goes down?

daniel-jodlos (changes requested):

lib/jellyfish/room_service.ex
LVala: You're absolutely correct, we also don't do kill Room GenServer and associated Socket processes on `RoomService.delete_room`, but that will be taken care of in another PR.

lib/jellyfish_web/socket.ex
LVala: But room can be deleted/crash after socket is initialised, but before some media event is received (I think so at least, or is that some nonsense?).

LVala:
I'm slightly uncertain where should I use what level of logging (warn, error etc.), I did it quite instinctively, so please pay attention to that.

test/jellyfish_web/socket_test.exs
```diff
@@ -0,0 +1,135 @@
...
+      assert {:ok, state} = connect(%{"room_id" => room_id, "peer_id" => peer_id})
+      assert state.peer_id == peer_id
+      assert state.room_id == room_id
+
+      status =
+        room_pid
+        |> Room.get_state()
+        |> Map.get(:peers)
+        |> Map.get(peer_id)
+        |> Map.get(:status)
```
mickel8, line 36: You might consider using `get_in`  but this depends on what we return from `Room.get_state`.

lib/jellyfish/room_service.ex
daniel-jodlos: I was actually referring to crashes. Normal exits will not activate the link. But I'm fine with handling them in another pull request

test/jellyfish_web/socket_test.exs
```diff
@@ -0,0 +1,135 @@
...
+      json = Jason.encode!(%{"notatype" => 45})
+
+      assert capture_log(fn -> send_from_client(json, room_id, peer_id) end) =~
+               "Received message with invalid structure"
+    end
+  end
+
+  describe "receiving messages from server" do
+    test "when it is a valid Media Event" do
+      assert {:push, {:text, json}, _state} = Socket.handle_info({:media_event, @data}, %{})
```
mickel8, line 105: nitpick but maybe we could have send_from_server (to be consistent with send_from_client) ?

lib/jellyfish/room.ex
```diff
@@ -106,32 +99,40 @@ defmodule Jellyfish.Room do
...
 
-        Process.monitor(socket_pid)
+          Process.monitor(socket_pid)
 
-        state =
-          state
-          |> put_in([:peers, peer_id, :socket_pid], socket_pid)
-          |> put_in([:peers, peer_id, :peer], %{peer_data.peer | status: :connected})
+          state =
+            put_in(state, [:peers, peer_id], %{
```
mickel8, line 122: IMO the previous version was more redable

mickel8 (changes requested):

bblaszkow06 (approved):

lib/jellyfish/room_service.ex
LVala: Yea, I mean that in general handling processes in some more proper way is not implemented.

lib/jellyfish/room.ex
```diff
@@ -44,38 +47,38 @@ defmodule Jellyfish.Room do
     GenServer.start_link(__MODULE__, args)
   end
 
-  @spec get_state(pid) :: t
+  @spec get_state(pid()) :: t()
```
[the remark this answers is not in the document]
LVala: I think it's ok to return the whole struct, it's not a part of public API anyways.

lib/jellyfish_web/socket.ex
```diff
@@ -19,36 +19,32 @@ defmodule JellyfishWeb.Socket do
     with {:ok, peer_id} <- Map.fetch(state.params, "peer_id"),
          {:ok, room_id} <- Map.fetch(state.params, "room_id"),
          {:ok, room_pid} <- RoomService.find_room(room_id),
+         :ok <- Logger.metadata(room_id: room_id),
```
mickel8, line 22: We shouldn't do that. `room_id` should only be included in logs of room process and all its children

mickel8 (changes requested):

lib/jellyfish_web/socket.ex
```diff
@@ -19,33 +19,41 @@ defmodule JellyfishWeb.Socket do
...
         state
+        |> Map.put(:room_id, room_id)
         |> Map.put(:peer_id, peer_id)
         |> Map.put(:room_pid, room_pid)
 
-      Logger.info("WebSocket connection from peer #{inspect(peer_id)} accepted")
+      Logger.info("""
+        WebSocket connection from peer #{inspect(peer_id)} accepted, \
+        room #{inspect(room_id)}
+      """)
```
mickel8, line 32: I am not sure whether these spaces are visible in logs but in general we write (at least in the engine):

```suggestion      
      Logger.info("""
      WebSocket connection from peer #{inspect(peer_id)} accepted, \
      room #{inspect(room_id)}
      """)
```

mickel8 (approved):

Rados13 (approved):

PR #16, opened 2023-03-03
Add cors plug

kamil-stasiak:


lib/jellyfish_web/endpoint.ex
```diff
@@ -47,5 +47,6 @@ defmodule JellyfishWeb.Endpoint do
   plug Plug.MethodOverride
   plug Plug.Head
   plug Plug.Session, @session_options
+  plug CORSPlug, origin: ["http://localhost:5173", "localhost:5173"]
```
mickel8, line 50: Just allow for every origin as we discussed and we can merge

mickel8 (approved):

PR #19, opened 2023-03-10
Phoenix 1.7

shuntrho:


LVala:
Read the docs carefully, views [still exist and are basically the same thing](https://hexdocs.pm/phoenix/controllers.html#new-rendering-formats), Phoenix docs actually still calls them views and components aren't replacement for them (you have even used them: `JellyfishWeb.ErrorJSON`, BTW its file name is called the other way than the auto generated one by Phoenix, here: `json_error.ex`, in new Phoenix project: `error_json.ex`). Moreover, they are now kept in the same directory as the controllers, naming conventions changed (e.g. `somethingJSON` in a file `controllers/something_json.ex`) and they don't need the `use ..., :view`. Function `json` that you used existed before Phoenix 1.7, and we have chosen to use views. Now, I think the amount of boilerplate is about the same, but it's a less granular (there's used to be one-to-one mapping between struct (e.g. `Peer`) and the view). But now, I don't think we can use functions like `render_many`, so the way it was before would require some tweaking besides moving files to different directories (see the new [Phoenix JSON and API guide](https://hexdocs.pm/phoenix/json_and_apis.html)). Honestly, I probably like the way Phoenix suggests doing it a bit more, even though it requires a bit more code, so IMO you didn't need to rewrite it. Please, have somebody else to also look at it and we can decide what to do.

I see that the PR is called `Phoenix 1.7`, so do you plan to use other nice features of 1.7 like Verified Routes (instead of auto-generated Phoenix functions for accessing routes in tests)? Also, I think it would be nice to re-generate the initial project boilerplate with `mix phx.new` like contents of `JellyfishWeb` module, as I noticed it contains some new things (like some stuff needed for verified routes, I think).

shuntrho:
Okay, in this case I think we should be using views as before.
Recreating project from template is a good idea and we may do that.

shuntrho:
Migrated to phoenix 1.7 properly, by creating new project under Phoenix 1.7.
Migrated to using verified routes and remained with using `Views`. Some tweaks were needed as I don't think we should be using `phoenix_views` dependency, so `render_one` and `render_many` are unavailable.

LVala (changes requested):

LVala (approved):

PR #21, opened 2023-03-16
Websocket API

shuntrho:
The HTML preview can be generated using asyncapi studio https://studio.asyncapi.com/ .

This is a very basic documentation of our websocket API. Its main purpose is to establish some common ground for how the messages should look like and discuss the API before implementing it.

In near future we will probably switch to using Protobufs, which will generate proper documentation for the messages.

### DIFF
--- a/docs/jellyfish-ws.yaml
+++ b/docs/jellyfish-ws.yaml
@@ -1,0 +1,95 @@
+asyncapi: '2.6.0'
+
+info:
+  title: Jellyfish Websocket API
+  version: '0.1.0'
+  contact:
+    name: MembraneTeam
+    url: https://membrane.stream/contact
+  description: |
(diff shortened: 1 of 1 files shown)


LVala (approved):
Seems good to me, although you can mention the "ping" messages, I'm not sure if Phoenix delivers some sort of "ping" functionality, or we have to implement it as e.g. `clientControlMessage`.

shuntrho:
It doesn't seem to be possible to send the `heartbeat` messages from the js API. We define our own heartbeat message and handle it on the server side.
Instead I propose to send the `ping` message from the server, it will be handled by the browser (no changes to the client required) and `pong` reply will be send`. The reply will be handled by the default implementation of the `handle_control` implementation of the `Socket.Transport`.

PR #22, opened 2023-03-16
Server auth

shuntrho:


lib/jellyfish_web/socket.ex
```diff
@@ -33,46 +37,97 @@ defmodule JellyfishWeb.Socket do
...
+
+        {:error, :no_params}
+
+      peer: {:ok, _status} ->
         Logger.warn("""
         WebSocket connection for peer #{inspect(state.params["peer_id"])} in room \
         #{inspect(state.params["room_id"])} already exists, rejected
         """)
 
         {:error, :already_connected}
```
Rados13, line 53: Won't we also go through this case if peer status is `:authenticating`? If yes I think the log should be adjusted.

lib/jellyfish_web/socket.ex
```diff
@@ -33,46 +37,97 @@ defmodule JellyfishWeb.Socket do
...
+    case Jason.decode(encoded_message) do
+      {:ok, %{"type" => "controlMessage", "data" => %{"type" => "authRequest", "token" => token}}} ->
+        with {:ok, peer_id} <-
+               Phoenix.Token.verify(
+                 JellyfishWeb.Endpoint,
+                 Application.get_env(:jellyfish, :auth_salt),
+                 token,
+                 max_age: Application.get_env(:jellyfish, :jwt_max_age)
+               ),
+             :ok <- if(peer_id == state.peer_id, do: :ok, else: {:error, :invalid_peer_id}) do
```
Rados13, line 96: What do you think about moving this `if` to a function like `verifey_peer_id` or something like that?

lib/jellyfish_web/socket.ex
```diff
@@ -33,46 +37,97 @@ defmodule JellyfishWeb.Socket do
...
+
+          {:push, {:text, message}, %{state | authenticated: true}}
+        else
+          {:error, reason} ->
+            Logger.warn("""
+            Authentication failed for peer #{inspect(state.peer_id)}, \
+            room: #{inspect(state.room_id)}. Reason: #{reason}
+            """)
+
+            {:push, {:text, unauthenticated_message()}, state}
```
Rados13, line 118: Wouldn't it be helpful to pass to the client a reason why authentication fails?

Rados13 (commented):
Also, what do you think about adding tests for scenarios where the config variable `server_auth` is set to false?

Rados13 (approved):

lib/jellyfish_web/controllers/peer_controller.ex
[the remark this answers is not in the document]
shuntrho: I think we might postpone it for later, perhaps to another PR. With tokens valid e.g. for a couple hours or a day it is rare that the peer will still be connected when the token expires.

Rados13 (commented):
I agree with the issues added by @LVala , so I remove the approval.

lib/jellyfish_web/token.ex
LVala, line 1: It's slightly bothering that the file has different name than the module (`token.ex` vs `PeerToken`)

test/jellyfish_web/socket_test.exs
```diff
@@ -1,102 +1,121 @@
...
-      status =
-        room_pid
-        |> Room.get_state()
-        |> Map.get(:peers)
-        |> Map.get(peer_id)
-        |> Map.get(:status)
-
-      assert status == :connected
+    test "when credentials are valid" do
+      assert {:ok, _state} = connect(%{})
```
LVala, line 35: I think this test is kinda useless now, at least it doesn't match the name `the credentials are valid` when there's no credentials anywhere.

LVala (approved):
SGTM, but please make sure to test it with updated sdk/dashboard, and maybe even have somebody else take a look at it just in case.

lib/jellyfish_web/socket.ex
```diff
@@ -14,65 +16,58 @@ defmodule JellyfishWeb.Socket do
...
+  end
 
-      {:error, :peer_not_found} ->
+  @impl true
+  def handle_in({encoded_message, [opcode: :text]}, %{authenticated?: false} = state) do
+    case Jason.decode(encoded_message) do
+      {:ok, %{"type" => "controlMessage", "data" => %{"type" => "authRequest", "token" => token}}} ->
+        with {:ok, %{peer_id: peer_id, room_id: room_id}} <- PeerToken.verify(token),
+             {:ok, room_pid} <- RoomService.find_room(room_id) do
+          Room.set_peer_connected(room_pid, peer_id)
```
LVala, line 35: Shouldn't we handle return value from this call to set_peer_connected? We probably don't want to have two WebSocket connections opened to the same peer.

LVala (changes requested):
Sorry, I noticed this only after the approve above, probably you need to take care of it.

Rados13 (approved):
Looks good, what do you think about adding a test for a situation, when a client tries to connect to a server with a valid token but a peer doesn't exist in this room for some reason?

lib/jellyfish_web/socket.ex
```diff
@@ -14,70 +16,58 @@ defmodule JellyfishWeb.Socket do
...
+            {:stop, :closed, {1000, inspect(reason)}, state}
+        end
+
+      _other ->
+        Logger.warn("""
+        Received message from unauthenticated peer, ignoring
+        """)
 
-    {:ok, state}
+        {:stop, :closed, {1000, "unauthenticated"}, state}
```
LVala, line 67: What does the `1000` mean here?

LVala (approved):
👍

PR #23, opened 2023-03-21
Rtc 178 improve handling room crashes

Rados13:


lib/jellyfish/room_service.ex
```diff
@@ -40,39 +55,63 @@ defmodule Jellyfish.RoomService do
...
+    {:noreply, state}
   end
 
   @impl true
-  def handle_call({:delete_room, _room_id}, _from, state) do
-    {:reply, {:error, :room_not_found}, state}
+  def handle_info({:DOWN, ref, :process, pid, reason}, state) do
+    Logger.warn("Process (#{inspect(ref)}, #{inspect(pid)}) is down with reason: #{reason}")
+
+    Phoenix.PubSub.broadcast(Jellyfish.PubSub, inspect(pid), :room_crashed)
```
bblaszkow06, line 101: If the room stops with reason `:normal` should we consider this to be a crash?

lib/jellyfish/room_service.ex
```diff
@@ -40,39 +55,63 @@ defmodule Jellyfish.RoomService do
...
+
+    {:noreply, state}
+  end
+
+  defp remove_room(room_id) do
+    case find_room(room_id) do
+      {:ok, pid} ->
+        true = Process.exit(pid, :kill)
```
bblaszkow06, line 109: Reason `:kill` will shut down the room without any cleanup - I don't think it should be the basic way to stop the room

lib/jellyfish_web/controllers/room_controller.ex
```diff
@@ -76,18 +80,18 @@ defmodule JellyfishWeb.RoomController do
...
 
   def show(conn, %{"room_id" => id}) do
-    case RoomService.find_room(id) do
-      {:ok, room_pid} ->
-        room =
-          room_pid
-          |> Room.get_state()
-          |> maps_to_lists()
+    with {:ok, room_pid} <- RoomService.find_room(id),
+         true <- Process.alive?(room_pid) do
```
bblaszkow06, line 84: This doesn't totally prevent calling a dead pid. IMO if you want to safely call `get_state` I'd add handling of `:noproc` error around `GenServer.call`

bblaszkow06 (changes requested):

lib/jellyfish/room_service.ex
```diff
@@ -15,17 +15,32 @@ defmodule Jellyfish.RoomService do
...
   end
 
   @spec list_rooms() :: [Room.t()]
   def list_rooms() do
-    :rooms
-    |> :ets.tab2list()
-    |> Enum.map(fn {_id, room_pid} -> Room.get_state(room_pid) end)
+    Jellyfish.RoomRegistry
+    |> Registry.select([{{:_, :"$1", :_}, [], [:"$1"]}])
+    |> Enum.filter(&Process.alive?(&1))
```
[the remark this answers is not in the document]
Rados13: According to Registry docs:
```
Looking up, dispatching and registering are efficient and immediate at the cost of delayed unsubscription. For example, if a process crashes, its keys are automatically removed from the registry but the change may not propagate immediately. This means certain operations may return processes that are already dead.
```
Based on this information I think that it is possible to get the dead process from the registry, but yes this line doesn't prevent a race condition, only minimalizes the possibility.
To sum up I will follow your suggestion from one of the following comments to properly handle `get_state`.

lib/jellyfish/room_service.ex
Rados13: I first tried using `Process.exit(pid,:normal)`, but this led to the strange behavior of the Registry. The registry didn't remove this process, so I decided to fix this by killing the process forcefully. But I didn't know that this would lead to a shutdown process without cleanup so I will try again using normal exit instead of kill.

lib/jellyfish/room.ex
```diff
@@ -47,9 +47,22 @@ defmodule Jellyfish.Room do
...
+
+  @spec get_state(pid()) :: t() | nil
   def get_state(room_pid) do
-    GenServer.call(room_pid, :state)
+    try do
+      GenServer.call(room_pid, :state)
+    catch
+      :exit, {:noproc, {GenServer, :call, [^room_pid, :state, _timeout]}} ->
+        Logger.warn(
+          "During get state of room #{inspect(room_pid)}, process exited because this process doesn't exist"
```
bblaszkow06, line 61: > process exited because this process doesn't exist

Umm, that's quite confusing :stuck_out_tongue: 
How about:
```suggestion
          "Cannot get state of #{inspect(room_pid)}, the room's process doesn't exist anymore"
```

lib/jellyfish/room_service.ex
```diff
@@ -40,39 +65,74 @@ defmodule Jellyfish.RoomService do
...
+            Logger.warn(
+              "During removing room #{room_id}, process exited because process didn't live already"
+            )
+        end
+
+      _not_found ->
+        Logger.warn("Room with id #{room_id} doesn't exist")
+    end
+
+    Logger.info("Deleted room #{inspect(room_id)}")
```
bblaszkow06, line 136: 1. `:via` tuple would simplify thigs here as well
2. I wouldn't log successful deletion if the process was dead already.

```suggestion
    try do
      :ok = GenServer.stop({:via, Registry, {Jellyfish.RoomRegistry, room_id}})
      Logger.info("Deleted room #{inspect(room_id)}")
    catch
      :exit, {:noproc, {GenServer, :stop, [_server, :normal, :infinity]}} ->
        Logger.warn("Room with id #{room_id} doesn't exist")
    end
    
    :ok
```

lib/jellyfish/room.ex
```diff
@@ -87,10 +100,18 @@ defmodule Jellyfish.Room do
...
 
     {:ok, state}
   end
 
+  @impl true
+  def terminate(reason, state) do
+    :ok = Registry.unregister(Jellyfish.RoomRegistry, state.id)
+    reason
+  end
+
```
bblaszkow06, line 114: It doesn't seem necessary.

Following up on this comment:

> (...) its keys are automatically removed from the registry but the change may not propagate immediately. This means certain operations may return processes that are already dead.
> (...)
> 
> Based on this information I think that it is possible to get the dead process from the registry, (...).

I've checked the implementation of `Registry`, and the removal of the registered process is based on a link (`register` calls `Process.link` with a partition pid and that process traps exits), so unregistration is only a matter of handling `:EXIT` message and removing entry from ETS in Registry process. Again, this doesn't fix a race condition, it may slightly reduce the chance of hitting it, so I'd get rid of this code

bblaszkow06 (changes requested):

PR #25, opened 2023-04-07
Set valid port in URL config, don't enforce https

bblaszkow06:
1. The port is set by env var, it should be matched in the URL setting
2. scheme will be inferred from the rest of the config. If not configurable by env vars, the default should be `http`

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -101,7 +101,7 @@ if config_env() == :prod do
   port = String.to_integer(System.get_env("PORT") || "4000")
 
   config :jellyfish, JellyfishWeb.Endpoint,
-    url: [host: host, port: 443, scheme: "https"],
+    url: [host: host, port: port],
     http: [
       # Enable IPv6 and bind on all interfaces.
       # Set it to  {0, 0, 0, 0, 0, 0, 0, 1} for local network only access.


Rados13 (approved):

mickel8:
Just want to mention that the previous config was auto-generated by Phoenix  and  `port` is used in `http` configuration a couple of lines below :thinking: 

I was trying to understand what `url` option is used for but I don't get it. Anyway, I assume that this is needed for running via Docker as we build our images with MIX_ENV=prod. 

If that's really needed, I think we can merge it

mickel8 (commented):
Just want to mention that the previous config was auto-generated by Phoenix and port is used in http configuration a couple of lines below :thinking: 

I was trying to understand what url option is used for but I don't get it. Anyway, I assume that this is needed for running via Docker as we build our images with MIX_ENV=prod.

If that's really needed, I think we can merge it

bblaszkow06:
Since the URL is only used in helpers of the endpoint (like [`url/0`](https://hexdocs.pm/phoenix/Phoenix.Endpoint.html#c:url/0)) and the only visible change is the URL printed on server start, this might not be a valid change.

PR #26, opened 2023-04-11
Use room id instead of room pid

mickel8:
This PR unifies API so that we use everywhere room id. I also changed topic names to room id and removed Room.start_link as it was not used.

### DIFF
--- a/lib/jellyfish_web/controllers/peer_controller.ex
+++ b/lib/jellyfish_web/controllers/peer_controller.ex
@@ -60,6 +60,10 @@ defmodule JellyfishWeb.PeerController do
     ]
 
   def create(conn, %{"room_id" => room_id} = params) do
+    # the room may crash between fetching its
+    # pid and adding a new peer to it
+    # in such a case, the controller will fail
+    # and Phoenix will return 500
     with {:ok, peer_type_string} <- Map.fetch(params, "type"),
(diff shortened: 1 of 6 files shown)


Rados13 (approved):

lib/jellyfish_web/controllers/peer_controller.ex
```diff
@@ -62,8 +62,8 @@ defmodule JellyfishWeb.PeerController do
   def create(conn, %{"room_id" => room_id} = params) do
     with {:ok, peer_type_string} <- Map.fetch(params, "type"),
          {:ok, peer_type} <- Peer.parse_type(peer_type_string),
-         {:ok, room_pid} <- RoomService.find_room(room_id),
-         {:ok, peer} <- Room.add_peer(room_pid, peer_type) do
+         {:ok, _room_pid} <- RoomService.find_room(room_id),
+         {:ok, peer} <- Room.add_peer(room_id, peer_type) do
```
bblaszkow06, line 66: This is a bit problematic as it doesn't ensure the process will be alive - it may shut down between `find_room` and `add_peer` calls. To properly handle that we'd have to wrap a call inside `add_peer`. I won't insist on changing that now, but I'd at least leave a note/todo here ;)

LVala (approved):
👍

PR #27, opened 2023-04-12
[RTC-208] Add basic server notifications

mickel8:
Let me add tests and docs in the next PR

### DIFF
--- a/lib/jellyfish_web/peer_socket.ex
+++ b/lib/jellyfish_web/peer_socket.ex
@@ -1,4 +1,4 @@
-defmodule JellyfishWeb.Socket do
+defmodule JellyfishWeb.PeerSocket do
   @moduledoc false
   @behaviour Phoenix.Socket.Transport
   require Logger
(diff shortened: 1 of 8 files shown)


lib/jellyfish_web/server_socket .ex
```diff
@@ -0,0 +1,125 @@
...
+
+  @impl true
+  def handle_info(:send_ping, state) do
+    Process.send_after(self(), :send_ping, @heartbeat_interval)
+    {:push, {:ping, ""}, state}
+  end
+
+  @impl true
+  def handle_info(msg, state) do
+    Logger.error(inspect(msg))
```
LVala, line 88: Is this intentional?

lib/jellyfish_web/peer_socket.ex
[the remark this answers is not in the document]
mickel8: Yeap, exactly

lib/jellyfish_web/server_socket .ex
mickel8: Fixed

PR #28, opened 2023-04-12
[RTC-192] RTSP component

sgfn:


lib/jellyfish/component/rtsp.ex
```diff
@@ -0,0 +1,27 @@
...
+
+  alias Membrane.RTC.Engine.Endpoint.RTSP
+
+  alias JellyfishWeb.ApiSpec
+
+  @impl true
+  def config(%{engine_pid: engine} = options) do
+    options = Map.drop(options, [:engine_pid, :room_id])
+
+    with {:ok, valid_opts} <- OpenApiSpex.cast_value(options, ApiSpec.Component.RTSP.schema()) do
```
sgfn, line 16: [The docs](https://hexdocs.pm/open_api_spex/3.16.1/readme.html#validating-and-casting-params) describe a way to enable global validation of requests, which can be done by plugging `OpenApiSpex.Plug.CastAndValidate` in the component controller.

I tried to use that option but couldn't get it to work. Will look more into it.

lib/jellyfish_web/api_spec/component.ex
```diff
@@ -18,18 +19,47 @@ defmodule JellyfishWeb.ApiSpec.Component do
...
-      example: %{
-        output_path: "/hls-output"
-      }
+      discriminator: %OpenApiSpex.Discriminator{
+        propertyName: "type",
+        mapping: %{
+          "hls" => HLS.Config,
+          "rtsp" => RTSP.Config
+        }
+      },
```
sgfn, line 58: Right now, the discriminator isn't used -- we might consider dropping it altogether

lib/jellyfish/component/rtsp.ex
```diff
@@ -0,0 +1,31 @@
...
+      component_spec =
+        Map.from_struct(valid_opts)
+        # OpenApiSpex will remove invalid options, so the following conversion, while ugly, is memory-safe
+        |> Map.new(fn {k, v} ->
+          {Atom.to_string(k) |> Macro.underscore() |> String.to_atom(), v}
```
mickel8, line 21: I am not sure if we can safely use Macro.underscore here :thinking: 

```
This function was designed to format language identifiers/tokens with the underscore-slash format, that's why it belongs to the [Macro](https://hexdocs.pm/elixir/Macro.html#content) module. Do not use it as a general mechanism for underscoring strings as it does not support Unicode or characters that are not valid in Elixir identifiers.
```

mix.exs
```diff
@@ -56,7 +56,7 @@ defmodule Jellyfish.MixProject do
       {:ymlr, "~> 3.0"},
 
       # Membrane deps
-      {:membrane_rtc_engine, github: "jellyfish-dev/membrane_rtc_engine"},
+      {:membrane_rtc_engine, github: "jellyfish-dev/membrane_rtc_engine", branch: "master"},
```
mickel8, line 59: master is default branch

test/jellyfish/component/rtsp_test.exs
```diff
@@ -0,0 +1,57 @@
...
+
+    expected =
+      Map.new(custom_opts, fn {k, v} -> {Macro.underscore(k) |> String.to_atom(), v} end)
+      |> Map.put(:rtc_engine, @engine_pid)
+      |> then(&struct(Endpoint.RTSP, &1))
+
+    {:ok, ^expected} = Component.RTSP.config(options)
+  end
+
+  test "error on no sourceUri" do
```
mickel8, line 41:
```suggestion
  test "missing required sourceUri" do
```

test/jellyfish_web/controllers/component_controller_test.exs
```diff
@@ -38,6 +38,26 @@ defmodule JellyfishWeb.ComponentControllerTest do
              } = json_response(conn, :ok)["data"]
     end
```
mickel8, line 40: Could we create two, separate `describe` blocks? One for hls tests and another for rtsp tests? Also, let's remove `@component_type` attribute from l:6

mickel8 (changes requested):

lib/jellyfish/component/rtsp.ex
sgfn: After consideration, we've decided to stick with the current approach when it comes to validation.

lib/jellyfish_web/api_spec/component.ex
sgfn: Dropped the discriminator

test/jellyfish_web/controllers/component_controller_test.exs
```diff
@@ -33,11 +32,31 @@ defmodule JellyfishWeb.ComponentControllerTest do
       assert %{
                "id" => ^room_id,
                "components" => [
-                 %{"id" => ^id, "type" => @component_type}
+                 %{"id" => ^id, "type" => "hls"}
                ]
              } = json_response(conn, :ok)["data"]
     end
 
+    test "renders errors when component type is invalid", %{conn: conn, room_id: room_id} do
```
mickel8: Move this one and the one from line 46 out of describe

mickel8 (approved):

LVala (approved):
👍

PR #45, opened 2023-06-27
[RTC-275] Allow for subscribing to notifications from a single room

LVala:


.gitmodules
```diff
@@ -1,5 +1,5 @@
 [submodule "protos"]
 	path = protos
 	url = https://github.com/jellyfish-dev/protos.git
-	branch = master
+	branch = per-room-notifications
```
LVala, line 4: This is going to be reverted when PR in `proto` repo is merged.

lib/jellyfish_web/server_socket.ex
```diff
@@ -73,10 +81,32 @@ defmodule JellyfishWeb.ServerSocket do
...
+         {:room_state_request, %RoomStateRequest{content: {_variant, option}}} <- request do
+      room_state = get_room_state(option)
+
+      reply =
+        %ServerMessage{content: room_state}
+        |> ServerMessage.encode()
+
+      {:reply, :ok, {:binary, reply}, state}
+    else
+      _other ->
```
bblaszkow06, line 95: In case of failure in the second clause, you'd probably want to inspect the unmatched value (`_other`), not the encoded message

bblaszkow06 (approved):

PR #46, opened 2023-06-27
Rtc 275 tcp turn fallback

roznawsk:
* Added TCP turn as fallback, removed tls (we have never used it anyway).
* Fixed simulcast.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -73,12 +73,9 @@ config :jellyfish,
   integrated_turn_port_range:
     System.get_env("INTEGRATED_TURN_PORT_RANGE", "50000-59999")
     |> ConfigParser.parse_integrated_turn_port_range(),
-  integrated_tcp_turn_port:
-    System.get_env("INTEGRATED_TCP_TURN_PORT")
-    |> ConfigParser.parse_port_number("INTEGRATED_TCP_TURN_PORT"),
-  integrated_tls_turn_port:
-    System.get_env("INTEGRATED_TLS_TURN_PORT")
-    |> ConfigParser.parse_port_number("INTEGRATED_TLS_TURN_PORT"),
+  integrated_turn_tcp_port:
+    System.get_env("INTEGRATED_TURN_TCP_PORT")
+    |> ConfigParser.parse_port_number("INTEGRATED_TURN_TCP_PORT"),
   integrated_turn_pkey: System.get_env("INTEGRATED_TURN_PKEY"),
   integrated_turn_cert: System.get_env("INTEGRATED_TURN_CERT"),
   integrated_turn_domain: System.get_env("VIRTUAL_HOST"),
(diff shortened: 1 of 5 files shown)


config/runtime.exs
```diff
@@ -73,12 +73,9 @@ config :jellyfish,
...
-    |> ConfigParser.parse_port_number("INTEGRATED_TCP_TURN_PORT"),
-  integrated_tls_turn_port:
-    System.get_env("INTEGRATED_TLS_TURN_PORT")
-    |> ConfigParser.parse_port_number("INTEGRATED_TLS_TURN_PORT"),
+  integrated_turn_tcp_port:
+    System.get_env("INTEGRATED_TURN_TCP_PORT")
+    |> ConfigParser.parse_port_number("INTEGRATED_TURN_TCP_PORT"),
   integrated_turn_pkey: System.get_env("INTEGRATED_TURN_PKEY"),
   integrated_turn_cert: System.get_env("INTEGRATED_TURN_CERT"),
   integrated_turn_domain: System.get_env("VIRTUAL_HOST"),
```
mickel8, line 81: Can be deleted too

lib/jellyfish/room.ex
```diff
@@ -312,32 +313,34 @@ defmodule Jellyfish.Room do
...
-
         turn_ip = Application.fetch_env!(:jellyfish, :integrated_turn_listen_ip)
         turn_mock_ip = Application.fetch_env!(:jellyfish, :integrated_turn_ip)
 
         integrated_turn_options = [
           ip: turn_ip,
           mock_ip: turn_mock_ip,
           ports_range: Application.fetch_env!(:jellyfish, :integrated_turn_port_range),
-          cert_file: turn_cert_file
+          cert_file: nil
```
mickel8, line 323: Just delete

lib/jellyfish/room.ex
```diff
@@ -312,32 +313,34 @@ defmodule Jellyfish.Room do
...
         ]
 
         [
           integrated_turn_options: integrated_turn_options,
-          integrated_turn_domain: Application.fetch_env!(:jellyfish, :integrated_turn_domain),
-          dtls_pkey: Application.get_env(:jellyfish, :dtls_pkey),
-          dtls_cert: Application.get_env(:jellyfish, :dtls_cert)
+          integrated_turn_domain: Application.fetch_env!(:jellyfish, :integrated_turn_domain)
```
mickel8, line 328: Can be deleted

lib/jellyfish/room.ex
```diff
@@ -312,32 +313,34 @@ defmodule Jellyfish.Room do
...
 
+    case Application.fetch_env!(:jellyfish, :integrated_turn_tcp_port) do
+      nil ->
+        :pass
+
+      tcp_turn_port ->
+        TURNManager.ensure_tcp_turn_launched(network_options[:integrated_turn_options],
+          port: tcp_turn_port
+        )
+    end
```
mickel8, line 342: I would favor

```elixir
tcp_turn_port = Application.fetch_env!()
if tcp_turn_port do
  TURNManager.ensure_tcp_turn_launched()
end
```

Also, if we use some function directly from the ICE plugin we have to include it in mix.exs

mickel8 (changes requested):

mickel8 (approved):

PR #47, opened 2023-07-05
Add subscription

roznawsk:
Receiving notifications requires subscribing to the topic first.
Ref:
- https://github.com/jellyfish-dev/elixir_server_sdk/pull/19
- https://github.com/jellyfish-dev/protos/pull/7

### DIFF
--- a/lib/jellyfish_web/server_socket.ex
+++ b/lib/jellyfish_web/server_socket.ex
@@ -88,24 +88,24 @@ defmodule JellyfishWeb.ServerSocket do
 
       {:reply, :ok, {:binary, reply}, state}
     else
+      {:error, request} ->
+        unexpected_message_error(request, state)
+
       other ->
-        Logger.warn("""
-        Received unexpected message on server WS.
-        Closing the connection.
-
-        Message: #{inspect(other)}
-        """)
-
-        {:stop, :closed, {1003, "operation not allowed"}, state}
+        unexpected_message_error(other, state)
     end
   end
 
   def handle_in({encoded_message, [opcode: _type]}, state) do
+    unexpected_message_error(encoded_message, state)
+  end
+
+  defp unexpected_message_error(msg, state) do
     Logger.warn("""
     Received unexpected message on server WS.
     Closing the connection.
 
-    Message: #{inspect(encoded_message)}
+    Message: #{inspect(msg)}
     """)
 
     {:stop, :closed, {1003, "operation not allowed"}, state}
@@ -131,7 +131,7 @@ defmodule JellyfishWeb.ServerSocket do
     {:ok, msg}
   end
 
-  defp handle_subscribe(request), do: request
+  defp handle_subscribe(request), do: {:error, request}
 
   @impl true
   def handle_info(:send_ping, state) do
--- a/lib/jellyfish_web/server_socket.ex
+++ b/lib/jellyfish_web/server_socket.ex
@@ -15,15 +15,11 @@ defmodule JellyfishWeb.ServerSocket do
     PeerCrashed,
     PeerDisconnected,
     RoomCrashed,
-    RoomNotFound,
-    RoomsState,
-    RoomState,
-    RoomStateRequest,
     SubscribeRequest,
     SubscriptionResponse
   }
 
-  alias Jellyfish.ServerMessage.RoomState.{Component, Config, Peer}
+  alias Jellyfish.ServerMessage.SubscriptionResponse.{RoomNotFound, RoomsState, RoomState}
 
   @heartbeat_interval 30_000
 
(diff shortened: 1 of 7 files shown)


lib/jellyfish_web/server_socket.ex
```diff
@@ -115,6 +110,27 @@ defmodule JellyfishWeb.ServerSocket do
...
+      :ok = Phoenix.PubSub.subscribe(Jellyfish.PubSub, event_type_to_topic(event))
+    end)
+
+    %ServerMessage{content: {:subscription_response, %SubscriptionResponse{}}}
+    |> ServerMessage.encode()
+    |> then(&{:ok, &1})
+  end
+
+  defp handle_request(request), do: request
+
```
LVala, line 133: I would remove the default case all together, or at least rise, we don't expect any other messages.

test/jellyfish_web/integration/server_socket_test.exs
```diff
@@ -217,4 +244,18 @@ defmodule JellyfishWeb.Integration.ServerSocketTest do
...
+    {room_id, _peer_id, peer_token, conn} = add_room_and_peer(conn, server_api_token)
+
+    {:ok, peer_ws} = WS.start_link("ws://127.0.0.1:#{@port}/socket/peer/websocket", :peer)
+    auth_request = peer_auth_request(peer_token)
+    :ok = WS.send_binary_frame(peer_ws, auth_request)
+
+    fn ->
+      conn = delete(conn, ~p"/room/#{room_id}/")
+      assert response(conn, :no_content)
+    end
```
LVala, line 259: This seems hacky, maybe let's create `on_exit` callback in `setup` block for the whole file which simply deletes all of the rooms. I considered doing this before but it wasn't necessary back then.

LVala (changes requested):

LVala:
I think the logic in `elixir_server_sdk` requirest some changes, see PR there.

lib/jellyfish_web/server_socket.ex
roznawsk: We do raise, in the `handle_in` callback.

In this clause we return `request` instead of `{:ok, mesage}` so we raise in the `with`.
However this isn't explicit really.

test/jellyfish_web/integration/server_socket_test.exs
roznawsk: Will do

roznawsk: Will do

lib/jellyfish_web/server_socket.ex
```diff
@@ -83,13 +84,7 @@ defmodule JellyfishWeb.ServerSocket do
 
   def handle_in({encoded_message, [opcode: :binary]}, state) do
```
sgfn, line 85:
```suggestion
  @impl true
  def handle_in({encoded_message, [opcode: :binary]}, state) do
```

sgfn: Nitpick; might as well annotate the `handle_in` callback from line 102.

lib/jellyfish_web/server_socket.ex
sgfn: I agree with @LVala on this -- I'd say either remove the default case, explicitly raise, or return something like `{:error, :unknown_request}` (and maybe log an error); either of these solutions is fine by me

test/jellyfish_web/integration/server_socket_test.exs
```diff
@@ -151,9 +169,17 @@ defmodule JellyfishWeb.Integration.ServerSocketTest do
     assert_receive %RoomNotFound{id: ^fake_room_id}
   end
 
+  test "doesn't send messages if not subscribed", %{conn: conn} do
```
sgfn, line 172: Not sure if we need to test for this, but we could consider adding a case like "crashes if attempted to subscribe to a nonexistent event type" -- I'd for sure want to test whether subscribing to multiple different event types works, but right now we have only one...

Could potentially just add a TODO for when another event type gets added, so that we don't forget about testing this functionality

roznawsk: We have a test, which checks whether sending an unexpected message causes websocket to close.
We cannot test whether subscribing to an nonexistent event causes such behavior, because this breaks on a lower level - Protobuf raises an error when trying to encode undefined event type.

lib/jellyfish_web/server_socket.ex
roznawsk: tbh, I prefer to make the annotations only once - above the first callback clause, but if everyone prefers adding annotations before each callback I will change this

test/jellyfish_web/integration/server_socket_test.exs
sgfn: Fair enough

lib/jellyfish_web/server_socket.ex
sgfn: I think it'd be clearer this way, GPT4 seems to agree, but ultimately it's your decision

sgfn (approved):
LGTM -- please wait for other reviews, though

LVala (approved):
👍, just little nitpicks

PR #48, opened 2023-07-07
add simple hls endpoint

blazpie:


Karolk99 (changes requested):
I read the whole PR but I would like to first address the problem with `hls_controller`.
I don't think that we should read the manifest from a file every time we get a manifest request. There are very frequent and reading from files is not that fast. Second thing, waiting for manifest using `PubSub` creates race condition (example at the end). I suggest creating a `GenServer` that will get all the updates from the `FileStorage` and `hls_controller` will send a cast message to this `GenServer` and wait for the answer. `GenServer` will send an answer immediately if it has the required answer otherwise it will try to send it when it gets an update from the `FileStorage`.

Race condition: 
Let's say that we have 0.5-second-long partial segments.
1) got a request for the manifest with a third partial of the third segment.
2) In the manifest file there is only a second partial of the third segment.
3) `PubSub` sends an update that there is a third partial of the third segment.
4) `hls_controller` subscribe and waits for the fourth partial of the third segment.

In the end, the request will take 0.5 seconds longer than it should.

PR #52, opened 2023-07-13
Add enforcing video codec for each room

Karolk99:


lib/jellyfish/room_service.ex
```diff
@@ -54,9 +54,10 @@ defmodule Jellyfish.RoomService do
     |> Enum.reject(&(&1 == nil))
   end
 
-  @spec create_room(Room.max_peers()) :: {:ok, Room.t()} | {:error, :bad_arg}
-  def create_room(max_peers) do
-    GenServer.call(__MODULE__, {:create_room, max_peers})
+  @spec create_room(Room.max_peers(), Room.enforce_encoding()) ::
+          {:ok, Room.t()} | {:error, :bad_arg, :max_peers | :enforceEncoding}
```
LVala, line 58:
```suggestion
          {:ok, Room.t()} | {:error, :invalid_max_args | :invalid_enforce_encoding}
```

lib/jellyfish_web/controllers/room_controller.ex
```diff
@@ -68,13 +68,18 @@ defmodule JellyfishWeb.RoomController do
...
       |> put_resp_content_type("application/json")
       |> put_status(:created)
       |> render("show.json", room: room)
     else
-      {:error, :bad_arg} -> {:error, :bad_request, "maxPeers must be a number"}
+      {:error, :bad_arg, :max_peers} ->
+        {:error, :bad_request, "maxPeers must be a number"}
+
+      {:error, :bad_arg, :enforce_encoding} ->
+        {:error, :bad_request, "enforceEncoding must be a h264 or vp8"}
```
LVala, line 82:
```suggestion
        {:error, :bad_request, "enforceEncoding must be 'h264' or 'vp8'"}
```

LVala (approved):

lib/jellyfish/room.ex
```diff
@@ -24,6 +24,7 @@ defmodule Jellyfish.Room do
 
   @type id :: String.t()
   @type max_peers :: non_neg_integer() | nil
+  @type enforce_encoding :: :h264 | :vp8 | nil
```
roznawsk, line 27:
```suggestion
  @type video_encoding :: :h264 | :vp8 | nil
```

LVala (approved):

lib/jellyfish/room_service.ex
```diff
@@ -54,9 +54,10 @@ defmodule Jellyfish.RoomService do
     |> Enum.reject(&(&1 == nil))
   end
 
-  @spec create_room(Room.max_peers()) :: {:ok, Room.t()} | {:error, :bad_arg}
-  def create_room(max_peers) do
-    GenServer.call(__MODULE__, {:create_room, max_peers})
+  @spec create_room(Room.max_peers(), Room.enforce_encoding()) ::
```
roznawsk, line 57: Is this spec valid? `Room.enforce_encoding` is an atom, and this function takes the encoding as a string.

lib/jellyfish/room_service.ex
```diff
@@ -70,24 +71,32 @@ defmodule Jellyfish.RoomService do
...
-  def handle_call({:create_room, max_peers}, _from, state)
-      when is_nil(max_peers) or (is_integer(max_peers) and max_peers >= 0) do
-    {:ok, room_pid, room_id} = Room.start(max_peers)
-    room = Room.get_state(room_id)
-    Process.monitor(room_pid)
+  def handle_call({:create_room, max_peers, enforce_encoding}, _from, state) do
+    with :ok <- validate_max_peers(max_peers),
+         {:ok, enforce_encoding} <- encoding_to_atom(enforce_encoding),
+         {:ok, room_pid, room_id} <- Room.start(max_peers, enforce_encoding) do
+      room = Room.get_state(room_id)
```
roznawsk, line 78: Room.start always returns ok, so we don't have to match.
```suggestion
         {:ok, enforce_encoding} <- encoding_to_atom(enforce_encoding) do
      {:ok, room_pid, room_id} = Room.start(max_peers, enforce_encoding)
      room = Room.get_state(room_id)
```

roznawsk (changes requested):
My main question is about the `enforceEncoding` name.

IMO it suggests whether encoding is enforced or not, while in fact it describes which encoding is enforced and it only applies to video encoding, while audio is always enforced to "OPUS".

So I think it would be more informative, if it was named `videoEncoding`.

PR #54, opened 2023-07-14
Add metrics

roznawsk:


lib/jellyfish_web/server_socket.ex
```diff
@@ -123,14 +126,16 @@ defmodule JellyfishWeb.ServerSocket do
...
+  defp handle_subscribe(%SubscribeRequest{
          id: id,
-         content: room_state
-       }}
+         event_type: {:metrics, %SubscribeRequest.Metrics{}}
+       }) do
+    :ok = Phoenix.PubSub.subscribe(Jellyfish.PubSub, "metrics")
 
-    {:ok, msg}
+    {:ok, %SubscriptionResponse{id: id}}
```
LVala, line 138: I see that we aren't passing the `content` field here, how does that work? Assuming that we are able to ignore this field, it seems little ambiguous, but I cannot think of a better alternative (one more field to `oneof` that is empty and means "response for metrics request"?, tough to tell)

lib/jellyfish/application.ex
```diff
@@ -7,18 +7,19 @@ defmodule Jellyfish.Application do
 
   @impl true
   def start(_type, _args) do
+    scrape_interval = Application.fetch_env!(:jellyfish, :metrics_scrape_interval)
```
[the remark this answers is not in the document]
roznawsk: You can't call `compile_env` inside a function, so I would have to create a module attribute with the value.
I want to avoid, as I had problems with fetching compile-time variables in the past.

lib/jellyfish/metrics_scraper.ex
```diff
@@ -0,0 +1,128 @@
...
+    :"ice.packets_sent"
+  ]
+
+  # {metric1, metric2, result_key}
+  # compound metrics are calculated as:
+  # (m1_value - old_m1_value) / (m2_value - old_m2_value)
+  # and saved under the `result_key`
+  @compound_metrics_to_derive [
+    {:"ice.buffers_processed_time", :"ice.buffers_processed", :avg_buff_proc_time}
+  ]
```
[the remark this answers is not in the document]
roznawsk: It's just copy pasted from `membrane_videoroom`. I think these metrics make sense as they are in use in videoroom.

lib/jellyfish_web/server_socket.ex
roznawsk: According to proto docs, oneof means zero or one of the options, so IMO it's appropriate to just not fill the content field of the message.

LVala: Fair enough

PR #57, opened 2023-07-20
Add simple load-balancing with use libcluster

Rados13:
[Description of approach](https://grand-ringer-10a.notion.site/Load-balancing-e7746f02575342bda97c937dc70904c8?pvs=4).
The third scenario is implemented. As for the metrics, the number of running rooms on the jellyfish instance is used for now. It will be changed to the number of forwarded tracks in the following PR.


lib/jellyfish/room_service.ex
```diff
@@ -55,9 +56,23 @@ defmodule Jellyfish.RoomService do
   end
 
   @spec create_room(Room.max_peers(), String.t()) ::
-          {:ok, Room.t()} | {:error, :invalid_max_peers | :invalid_video_codec}
+          {:ok, Room.t(), String.t()} | {:error, :invalid_max_peers | :invalid_video_codec}
   def create_room(max_peers, video_codec) do
-    GenServer.call(__MODULE__, {:create_room, max_peers, video_codec})
+    :ok = Phoenix.PubSub.broadcast(Jellyfish.PubSub, "jellyfishes", {:resource_usage, self()})
```
roznawsk, line 61:
```suggestion
    :ok = Phoenix.PubSub.broadcast(Jellyfish.PubSub, "jellyfishes", {:get_resource_usage, self()})
```

roznawsk (changes requested):
I have a general comment regarding the strategy for getting resource usage.

I wonder whether we should always ask all the jellyfishes about their resource usage every time we create a room. With a couple jellyfishes and traffic the requests to create a room can take some time.
Instead we could consider regularly sending resource report, so that we have a candidate jellyfish to create a room before we even receive the request. However, this solution probably comes with its disadvantages as well.
Overall, I think we can discuss the topic outside this PR.

lib/jellyfish/room_service.ex
```diff
@@ -137,6 +170,40 @@ defmodule Jellyfish.RoomService do
...
 
+  defp receive_resources(results \\ [], nodes \\ []) do
+    receive do
+      {:resources, node_name, resource_usage} = msg ->
+        new_results = [resource_usage | results]
+        new_nodes = [node_name | nodes]
+
+        cond do
+          node_name in nodes ->
+            Process.send_after(self(), msg, 500)
```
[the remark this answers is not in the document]
Rados13: Ok, I made here mistake. I thought that there was a possibility that we could receive messages from two different broadcasted messages.

lib/jellyfish/room_service.ex
```diff
@@ -137,6 +170,34 @@ defmodule Jellyfish.RoomService do
...
+        Logger.warn(
+          "Nodes that don't respond on resource usage request #{inspect(unknown_nodes)}"
+        )
+
+        Enum.zip(nodes, results)
+    end
+  end
+
+  defp get_resource_usage() do
+    RoomService.list_rooms() |> Enum.count()
```
roznawsk, line 198: It's probably going to change in the next PR, but the `RoomService.list_rooms()` function gets state of every room by making a blocking call to each of them. We only need the number of rooms so we could add a new function to `RoomService` which only gets the room ids from the registry. (without getting their state)

If you're going to change it anyway you can ignore this comment.

roznawsk (approved):
LGTM

lib/jellyfish/room_service.ex
Rados13: Yup, plan is to change it in the following PR because the metric of the number of rooms is very naive.

lib/jellyfish/room_service.ex
```diff
@@ -55,9 +56,23 @@ defmodule Jellyfish.RoomService do
...
   def create_room(max_peers, video_codec) do
-    GenServer.call(__MODULE__, {:create_room, max_peers, video_codec})
+    :ok = Phoenix.PubSub.broadcast(Jellyfish.PubSub, "jellyfishes", {:get_resource_usage, self()})
+
+    node_resources = receive_resources()
+
+    {min_node, _room_size} =
+      Enum.min_by(node_resources, fn {_node_name, room_num} -> room_num end)
+
+    Logger.info(node_resources)
```
sgfn, line 68: Nitpick but we could make this log here more descriptive

config/runtime.exs
```diff
@@ -63,6 +63,22 @@ defmodule ConfigParser do
   end
 end
 
+hosts =
+  (System.get_env("NODES", nil) || "")
+  |> String.split(" ")
+  |> Enum.reject(&(&1 == ""))
+  |> Enum.map(&String.to_atom(&1))
+
+IO.inspect(hosts, label: :hosts)
```
LVala, line 72: Is this intentional?

docker-compose.yaml
LVala, line 1: These 2 `docker compose` services looks almost identical, I was wondering if you could somehow make 1 generic service and extend it with needed variables like `PORT`, `NODE_NAME` etc. [Here](https://stackoverflow.com/questions/45279662/extend-service-in-docker-compose-3)'s some info about that (not sure if entirely relevant, I only took a very quick peek at it).

test/jellyfish/cluster/load_balancing_test.exs
```diff
@@ -0,0 +1,72 @@
...
+  end
+
+  if Mix.env() == :test do
+    defp map_jellyfish_address(jellyfish), do: jellyfish
+  else
+    defp map_jellyfish_address(jellyfish) do
+      %{
+        @node1 => "app1:4001",
+        @node2 => "app2:4002"
+      }[jellyfish]
```
sgfn, line 58: Nitpick, but I got seriously confused by this syntax 😆
I'd personally use `Map.get` here, but it's up to you

lib/jellyfish/application.ex
```diff
@@ -19,7 +20,8 @@ defmodule Jellyfish.Application do
       JellyfishWeb.Endpoint,
       # Start the RoomService
       Jellyfish.RoomService,
-      {Registry, keys: :unique, name: Jellyfish.RoomRegistry}
+      {Registry, keys: :unique, name: Jellyfish.RoomRegistry},
+      {Cluster.Supervisor, [topologies, [name: Jellyfish.ClusterSupervisor]]}
```
sgfn, line 24: Just curious -- does this mean that we'll always want to start the `Cluster.Supervisor`, even when not actually using distribution?

sgfn (changes requested):

lib/jellyfish/application.ex
Rados13: If I understand correctly how Supervisor in Beam works, then yes.

lib/jellyfish/room_service.ex
Rados13: This change is made in #59

sgfn (changes requested):
One small issue

LVala (approved):

roznawsk (approved):

roznawsk (approved):

sgfn (approved):
LGTM :seal:

docker-compose.yaml
```diff
@@ -0,0 +1,69 @@
...
+      - sh
+      - -c
+      - |
+        cd app/ 
+        mix deps.get 
+        MIX_ENV=test_containerised mix test --only containerised
+    volumes:
+      - .:/app
+      - /app/_build
+      - /app/deps
```
mickel8, line 40: Why those volumes?

mickel8 (changes requested):

lib/jellyfish/room_service.ex
```diff
@@ -89,7 +107,7 @@ defmodule Jellyfish.RoomService do
         {:room_created, room_id}
       )
 
-      {:reply, {:ok, room}, state}
+      {:reply, {:ok, room, Application.fetch_env!(:jellyfish, :address)}, state}
```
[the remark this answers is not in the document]
Rados13: The caller knows another node by its node name not by its address and port on which is listening. Also, we have to somehow send to which jellyfish it should connect.

docker-compose.yaml
Rados13: As we run it on `docker_membrane,` project files are not included. Then we exclude `_build` and `deps` directory because it could be compiled with different flag etc.

lib/jellyfish/room_service.ex
```diff
@@ -137,6 +167,34 @@ defmodule Jellyfish.RoomService do
     {:noreply, state}
   end
 
+  defp receive_resources(results \\ [], nodes \\ []) do
```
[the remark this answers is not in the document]
Rados13: But multicall doesn't solve a problem that process could die in the meantime. If you prefer it to be multicall I can change. 
As for RoomService being a GenServer if I remember correctly it was done to avoid a situation where a room would be created two times by different users, but I am not sure if this is still valid.

lib/jellyfish/room_service.ex
```diff
@@ -57,9 +57,12 @@ defmodule Jellyfish.RoomService do
   @spec create_room(Room.max_peers(), String.t()) ::
           {:ok, Room.t(), String.t()} | {:error, :invalid_max_peers | :invalid_video_codec}
   def create_room(max_peers, video_codec) do
-    :ok = Phoenix.PubSub.broadcast(Jellyfish.PubSub, "jellyfishes", {:get_resource_usage, self()})
+    {node_resources, failed_nodes} =
+      :rpc.multicall(Jellyfish.RoomService, :get_resource_usage, [])
```
mickel8, line 61: In theory, node_resources might contain tuples `{:badrpc, Reason}`. Not sure if we should filter and log them?

lib/jellyfish/room_service.ex
```diff
@@ -57,9 +57,12 @@ defmodule Jellyfish.RoomService do
...
 
-    node_resources = receive_resources()
+    if Enum.count(failed_nodes) > 0 do
+      Logger.warn("Nodes that don't respond on resource usage request #{inspect(failed_nodes)}")
```
mickel8, line 64:
```suggestion
      Logger.warn("Couldn't get resource usage of the following nodes. Reason: nodes don't exist. Nodes: #{inspect(failed_nodes)}")
```

lib/jellyfish/room_service.ex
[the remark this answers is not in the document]
mickel8: I meant that multicall returns a list of bad nodes out of the box. We don't have to manually wait x seconds as in the previous solution :wink:

test/jellyfish/cluster/load_balancing_test.exs
```diff
@@ -18,27 +18,35 @@ defmodule Jellyfish.Cluster.LoadBalancingTest do
...
+    assert_rooms_number_on_jellyfish(jellyfish_instance2, 1)
 
     assert_rooms_number_on_jellyfish(node1, 1)
     assert_rooms_number_on_jellyfish(node2, 1)
+
+    room_id = response_body1 |> Jason.decode!() |> get_in(["data", "id"])
+
+    delete_room(jellyfish_instance1, room_id)
+
+    assert_rooms_number_on_jellyfish(jellyfish_instance1, 0)
```
mickel8, line 48: And now create a room and make sure that we still have single room on every machine :stuck_out_tongue:

test/jellyfish/cluster/load_balancing_test.exs
```diff
@@ -18,27 +18,35 @@ defmodule Jellyfish.Cluster.LoadBalancingTest do
   @tag timeout: @max_test_duration
   test "spawning tasks on a cluster" do
     [node1, node2] =
-      if Mix.env() == :test do
+      if Mix.env() == :ci do
+        # On CI we don't use Divo, because we don't want to run Docker in Docker
+        ["app1:4001", "app2:4002"]
+      else
         Divo.Suite.start(services: [:app1, :app2]) |> on_exit()
```
mickel8, line 25: I forgot to mention but Divo.Suite.start is a private function, isn't it?

mix.exs
```diff
@@ -100,8 +101,8 @@ defmodule Jellyfish.MixProject do
       setup: ["deps.get"],
       "api.spec": ["openapi.spec.yaml --spec JellyfishWeb.ApiSpec"],
       test: ["test --exclude containerised"],
-      test_containerised: ["test --only containerised"],
-      test_containerised_compose: ["cmd docker compose run test; docker compose stop"]
+      "test.cluster": ["test --only containerised"],
```
mickel8, line 104:
```suggestion
      "test.cluster": ["test --only cluster"],
```

mix.exs
```diff
@@ -100,8 +101,8 @@ defmodule Jellyfish.MixProject do
...
+      "test.cluster.compose": ["cmd docker compose run test; docker compose stop"]
```
mickel8, line 105: actually test.cluster.ci should be more meaningful

openapi.yaml
```diff
@@ -203,6 +203,22 @@ components:
...
       x-struct: Elixir.JellyfishWeb.ApiSpec.Room.Config
+    RoomCreateDetailsResponse:
+      description: Response containing room details
+      properties:
+        data:
+          properties:
+            room: 
+              $ref: '#/components/schemas/Room'
+            jellyfish_address:
+              description: Jellyfish instance address
```
mickel8, line 214:
```suggestion
              description: Jellyfish instance address where the room was created. This might be different than the address of Jellyfish where the request was sent only when running a cluster of Jellyfishes. 
```

lib/jellyfish/room_service.ex
Rados13: Yup, I found out about that in #59. After fixing it there, I will also add this fix here.

mickel8 (approved):

lib/jellyfish/room_service.ex
mickel8: kk, we can merge this without waiting for #59 and fix this in #59

PR #58, opened 2023-07-20
Add hls playable field and notification

Karolk99:


test/jellyfish_web/integration/server_socket_test.exs
```diff
@@ -175,6 +184,21 @@ defmodule JellyfishWeb.Integration.ServerSocketTest do
...
 
+    correct_components? =
+      components
+      |> Enum.map(fn %Component{component: component} -> component end)
+      |> Enum.all?(fn
+        {:hls, %Hls{playable: false}} -> true
+        {:rtsp, %Rtsp{}} -> true
+      end)
+
+    assert correct_components?
```
Rados13, line 195:
```suggestion
    assert true =
      components
      |> Enum.map(fn %Component{component: component} -> component end)
      |> Enum.all?(fn
        {:hls, %Hls{playable: false}} -> true
        {:rtsp, %Rtsp{}} -> true
        _other -> false
      end)

```

lib/jellyfish/room.ex
```diff
@@ -321,6 +322,25 @@ defmodule Jellyfish.Room do
...
+      end)
+
+    Phoenix.PubSub.broadcast(
+      Jellyfish.PubSub,
+      "server_notification",
+      {:hls_playable, state.id, endpoint_id}
+    )
+
+    {:noreply, state}
+  end
```
Rados13, line 342: Shouldn't we store this information somewhere?

lib/jellyfish/component.ex
```diff
@@ -44,14 +48,20 @@ defmodule Jellyfish.Component do
   @spec new(component(), map()) :: {:ok, t()} | {:error, term()}
   def new(type, options) do
     with {:ok, endpoint} <- type.config(options) do
+      metadata = get_metadata(type)
```
roznawsk, line 51: Perhaps we could move the logic of getting metadata to the modules in question?
```suggestion
      metadata = type.metadata()
```

lib/jellyfish/room.ex
roznawsk: I think it should be stored inside the HLS component

roznawsk (changes requested):

LVala (approved):
👍

PR #59, opened 2023-07-24
Use number of forwarded tracks instead of number of rooms in loadbalancing

Rados13:


lib/jellyfish/room_service.ex
```diff
@@ -192,7 +191,38 @@ defmodule Jellyfish.RoomService do
...
+  defp accumulate_resource_responses(room_ids, acc \\ []) do
+    receive do
+      response ->
+        acc = [response | acc]
+
+        if Enum.count(acc) == Enum.count(room_ids) do
+          acc
+        else
+          accumulate_resource_responses(room_ids, acc)
+        end
```
LVala, line 213: If I was very pedantic (I'm not) I'd say you could add simply the number of forwarded tracks to the `acc` insteady of tuple `{room_id, forwarded_tracks}` and simply `Enum.sum/1` instad of the `Enum.reduce/3`.

LVala (approved):
👍

lib/jellyfish/room_service.ex
```diff
@@ -192,7 +191,38 @@ defmodule Jellyfish.RoomService do
...
+    after
+      500 ->
```
roznawsk, line 215: Maybe put this timeout value in a module attribute?

lib/jellyfish/room.ex
```diff
@@ -72,6 +72,11 @@ defmodule Jellyfish.Room do
     end
   end
 
+  @spec get_forwarded_tracks_in_room(id()) :: :ok
```
roznawsk, line 75:
```suggestion
  @spec request_forwarded_tracks_in_room(id()) :: :ok
```

Maybe change the name somehow, so it doesn't sound like this functions returns the forwarded tracks?

roznawsk (changes requested):

roznawsk (approved):

lib/jellyfish/room.ex
```diff
@@ -72,6 +72,11 @@ defmodule Jellyfish.Room do
     end
   end
 
+  @spec request_forwarded_tracks_number_in_room(id()) :: :ok
+  def request_forwarded_tracks_number_in_room(room_id) do
+    GenServer.cast(registry_id(room_id), {:forwarded_tracks_number, self()})
```
sgfn, line 77: Same thing as with the function name in the engine -- but that's a nitpick

sgfn (approved):

lib/jellyfish/room_service.ex
```diff
@@ -80,9 +102,21 @@ defmodule Jellyfish.RoomService do
...
+  @spec get_resource_usage() :: %{
+          node: Node.t(),
+          forwarded_tracks_number: integer(),
+          rooms_number: integer()
+        }
   def get_resource_usage() do
-    list_rooms() |> Enum.count() |> then(&{Node.self(), &1})
+    room_ids = get_rooms_ids()
+
+    Enum.each(room_ids, &Room.request_forwarded_tracks_number_in_room(&1))
```
mickel8, line 113: I get the idea but I think it will be easier to make `request_forwarded_tracks_number_in_room` synchronous and spawn multiple tasks to get it in a concurrent way. You can use Task.away_many

lib/jellyfish/room.ex
mickel8: I would call this function the same way it's called in rtc engine

mickel8 (changes requested):

lib/jellyfish/room_service.ex
```diff
@@ -66,8 +65,29 @@ defmodule Jellyfish.RoomService do
...
+        fn
+          %{forwarded_tracks_number: forwarded_tracks, rooms_number: rooms_num1},
+          %{forwarded_tracks_number: forwarded_tracks, rooms_number: rooms_num2} ->
+            rooms_num1 < rooms_num2
+
+          %{forwarded_tracks_number: forwarded_tracks1},
+          %{forwarded_tracks_number: forwarded_tracks2} ->
+            forwarded_tracks1 < forwarded_tracks2
+        end
+      )
```
mickel8, line 90: Let's put this in a separate function called e.g. `find_best_node` that will return single node

mickel8 (approved):

sgfn (approved):
Go ahead

PR #62, opened 2023-07-31
[RTC-310] Prometheus metrics for Jellyfish

sgfn:
The existing (and new) metric names were changed according to the [Prometheus guidelines](https://prometheus.io/docs/practices/naming/) (see also [this page](https://hexdocs.pm/telemetry_metrics_prometheus_core/1.0.0/TelemetryMetricsPrometheus.Core.html#module-naming) from the `telemetry_metrics_prometheus_core` docs).

Explanation for the Phoenix metrics can be found [here](https://hexdocs.pm/phoenix/Phoenix.Logger.html). Personally, I don't really see any use for the metrics which use `system_time` measurements (they return the system time of when a given event occured, so their graph in time is a linear function ¯\_(ツ)_/¯), so I think we should drop them -- but we can discuss this.

EDIT: Right now, as we're subscribed to ICE events, the traffic metrics **DO NOT INCLUDE** traffic from/to RTSP and HLS components. The Cloud team has agreed to this (right now, they're going to use these metrics as a PoC anyway), but we have to keep this in mind for the future.


lib/jellyfish_web/telemetry.ex
```diff
@@ -11,61 +12,73 @@ defmodule JellyfishWeb.Telemetry do
...
+      {TelemetryMetricsPrometheus, metrics: metrics(&last_value/2)}
     ]
 
     Supervisor.init(children, strategy: :one_for_one)
   end
 
-  def metrics do
+  # Phoenix by default uses the `summary` metric type in `LiveDashboard`,
+  # but `TelemetryMetricsPrometheus` doesn't support it, so we have to use `last_value` instead
+  def metrics(metric_type \\ &summary/2) do
```
mickel8, line 24: Could we please link Prometheus and TelemetryMetricsPrometheus guides on naming events, measurements etc.?

lib/jellyfish_web/telemetry/metrics_aggregator.ex
```diff
@@ -0,0 +1,116 @@
...
+  @ice_received_event [Membrane.ICE, :ice, :payload, :received]
+  @ice_sent_event [Membrane.ICE, :ice, :payload, :sent]
+
+  @spec start_link(Keyword.t()) :: GenServer.on_start()
+  def start_link(opts),
+    do: GenServer.start_link(__MODULE__, opts, name: __MODULE__)
+
+  @spec handle_event(term(), map(), map(), map()) :: :ok
+  def handle_event(name, measurements, metadata, config),
+    do: GenServer.cast(config.handler, {:event, name, measurements, metadata})
```
mickel8, line 18: We will have a lot of ice events so instead of sending messages better use ets

lib/jellyfish_web/telemetry/metrics_aggregator.ex
```diff
@@ -0,0 +1,116 @@
...
+      ),
+      last_value("jellyfish.rooms"),
+      last_value("jellyfish.room.peers",
+        tags: [:room_id]
+      ),
+      sum("jellyfish.room.peer_time.total.seconds",
+        event_name: [:jellyfish, :room],
+        measurement: :peer_time_total,
+        tags: [:room_id]
+      )
```
mickel8, line 49: That might turn out to be problematic in the future as prometheus warns about high cardinality sets of labels https://prometheus.io/docs/practices/naming/#labels

mickel8 (changes requested):
> I don't really see any use for the metrics which use system_time measurements (they return the system time of when a given event occured, so their graph in time is a linear function ¯_(ツ)_/¯), so I think we should drop them -- but we can discuss this

I don't see any usage of those metrics too but Phoenix Team wouldn't include something that is useless, would it?

sgfn:
> I don't see any usage of those metrics too but Phoenix Team wouldn't include something that is useless, would it?

Cloud says they don't need them, as every metric update in Prometheus comes with a timestamp, anyway. But it doesn't hurt to keep them, I guess

lib/jellyfish_web/telemetry/metrics_aggregator.ex
sgfn: Yeah, for now I'll add a comment warning about this.

If I get any info back from Cloud on this issue before merging this, I'll post it here

mickel8 (approved):

PR #65, opened 2023-08-07
Implement hls storage

Karolk99:
This PR is part of a larger task of adding HLS to Jellyfish. In this PR there is an implementation of HLS Storage which is the only process that communicates with HLS Endpoint.

Here is a short introduction [doc](https://membraneframework.atlassian.net/wiki/spaces/MEMBRANEFR/pages/edit-v2/118063105?draftShareId=00bcc812-bfe6-4d45-abdb-7b264a91eea1&inEditorTemplatesPanel=auto_closed).

### DIFF
--- a/lib/jellyfish/component/hls/ets_helper.ex
+++ b/lib/jellyfish/component/hls/ets_helper.ex
@@ -6,93 +6,17 @@ defmodule Jellyfish.Component.HLS.EtsHelper do
   @rooms_to_tabels :rooms_to_tables
   @free_tables :free_tables
 
-  @last_partial_key :last_partial
+  @recent_partial_key :recent_partial
   @manifest_key :manifest
 
-  @delta_last_partial_key :delta_last_partial
+  @delta_recent_partial_key :delta_recent_partial
   @delta_manifest_key :delta_manifest
 
   @type partial :: {non_neg_integer(), non_neg_integer()}
 
-  def get_partial(room_id, filename, offset) do
-    table = get_table(room_id)
-    key = generate_partial_key(filename, offset)
-    [{^key, partial}] = :ets.lookup(table, key)
-    {:ok, partial}
-  end
-
-  def get_last_partial(room_id, filename) do
-    table = get_table(room_id)
-
-    [{_key, last_partial}] =
-      if String.contains?(filename, "_delta.m3u8"),
-        do: :ets.lookup(table, @delta_last_partial_key),
-        else: :ets.lookup(table, @last_partial_key)
-
-    last_partial
-  end
-
-  def get_manifest(room_id, filename) do
-    table = get_table(room_id)
-
-    [{_key, manifest}] =
-      if String.contains?(filename, "_delta.m3u8"),
-        do: :ets.lookup(table, @delta_manifest_key),
-        else: :ets.lookup(table, @manifest_key)
-
-    manifest
-  end
-
-  @spec add_last_partial(Room.id(), partial()) :: true
-  def add_last_partial(room_id, last_partial) do
-    table = get_table(room_id)
-    :ets.insert(table, {@last_partial_key, last_partial})
-  end
-
-  @spec add_last_partial(Room.id(), partial()) :: true
-  def add_delta_last_partial(room_id, delta_last_partial) do
-    table = get_table(room_id)
-    :ets.insert(table, {@delta_last_partial_key, delta_last_partial})
-  end
-
-  @spec add_delta_manifest(Room.id(), String.t()) :: true
-  def add_delta_manifest(room_id, delta_manifest) do
-    table = get_table(room_id)
-    :ets.insert(table, {@delta_manifest_key, delta_manifest})
-  end
-
-  @spec add_manifest(Room.id(), String.t()) :: true
-  def add_manifest(room_id, manifest) do
-    table = get_table(room_id)
-    :ets.insert(table, {@manifest_key, manifest})
-  end
-
-  @spec remove_partial(Room.id(), String.t(), non_neg_integer()) :: true
-  def remove_partial(room_id, filename, offset) do
-    table = get_table(room_id)
-    key = generate_partial_key(filename, offset)
-    :ets.delete(table, key)
-  end
-
-  @spec add_partial(Room.id(), binary(), String.t(), non_neg_integer()) :: true
-  def add_partial(room_id, partial, filename, offset) do
-    table = get_table(room_id)
-    key = generate_partial_key(filename, offset)
-    :ets.insert(table, {key, partial})
-  end
-
-  @spec remove_room(Room.id()) :: :ok | {:error, :not_exists}
-  def remove_room(room_id) do
-    case :ets.lookup(@rooms_to_tabels, room_id) do
-      [{^room_id, table}] ->
-        add_to_free_tables(table)
-        :ets.delete(@rooms_to_tabels, room_id)
-        :ok
-
-      _empty ->
-        {:error, :not_exists}
-    end
-  end
+  ###
+  ### ROOM MANAGMENT
+  ###
 
   @spec add_room(Room.id()) :: :ok | {:error, :already_exists}
   def add_room(room_id) do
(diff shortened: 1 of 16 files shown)


lib/jellyfish/component/hls/storage.ex
```diff
@@ -0,0 +1,201 @@
...
+  @behaviour Membrane.HTTPAdaptiveStream.Storage
+
+  alias Jellyfish.Room
+
+  @enforce_keys [:directory, :room_id]
+  defstruct @enforce_keys ++ [partial_sn: nil, segment_sn: 0, partials_in_ets: []]
+
+  @type partial_ets_key :: String.t()
+  @type segment_sn :: non_neg_integer()
+  @type partial_sn :: non_neg_integer()
```
Rados13, line 13:
```suggestion
  @type partial_sn :: non_neg_integer() | nil
```

Rados13: Or one type for both segments and partials.
```elixir
@type sequence_number :: non_neg_integer()
```

lib/jellyfish/component/hls/storage.ex
```diff
@@ -0,0 +1,201 @@
...
+
+  @manifest_key :manifest
+  @last_partial_key :last_partial
+
+  @ets_duration_in_segments 4
+
+  @impl true
+  def init(%{directory: directory, room_id: room_id}) do
+    # Initializes ets storage that will serve partial segments and manifests
+    :ets.new(room_id, [:public, :set, :named_table])
```
Rados13, line 32: What do you think about using for ETS storage names like `{__MODULE__, room_id}` or something like this? This way we reduce the possibility that any other module would use ets storage with the same name.

lib/jellyfish/component/hls/storage.ex
```diff
@@ -0,0 +1,201 @@
...
+
+    %__MODULE__{room_id: room_id, directory: directory}
+  end
+
+  @impl true
+  def store(_perent_id, name, content, metadata, context, state) do
```
Rados13, line 38:
```suggestion
  def store(_parent_id, name, content, metadata, context, state) do
```

lib/jellyfish/component/hls/storage.ex
```diff
@@ -0,0 +1,201 @@
...
+
+  defp store_segment(
+         filename,
+         content,
+         %{directory: directory} = state
+       ) do
+    result =
+      directory
+      |> Path.join(filename)
+      |> File.write(content, [:binary])
```
Rados13, line 72: What do you think about creating a util function for writing to file?
```elixir
defp write_to_file(directory, filename, content, write_options) do 
      directory
      |> Path.join(filename)
      |> File.write(content, write_options)
end

File: lib/jellyfish/component/hls/storage.ex [hunk 5]
```diff
@@ -186,8 +178,8 @@ defmodule Jellyfish.Component.HLS.Storage do
          partial_sn: partial_sn,
          room_id: room_id
        }) do
-    :ets.insert(room_id, {@manifest_key, manifest})
-    :ets.insert(room_id, {@last_partial_key, {segment_sn, partial_sn}})
+    :ets.insert({__MODULE__, room_id}, {@manifest_key, manifest})
+    :ets.insert({__MODULE__, room_id}, {@last_partial_key, {segment_sn, partial_sn}})
```
Rados13, line 182: As we are using `{__MODULE__, room_id}` in multiple places I think we should a helper function for creating a ets table name.

lib/jellyfish/component/hls/ets_helper.ex
```diff
@@ -0,0 +1,192 @@
...
+    end
+  end
+
+  defp add_to_free_tables(table) do
+    [{@free_tables, free_tables}] = :ets.lookup(@rooms_to_tabels, @free_tables)
+    :ets.insert(@rooms_to_tabels, {@free_tables, [table | free_tables]})
+  end
+
+  defp room_exists?(room_id) do
+    if :ets.lookup(@rooms_to_tabels, room_id) == [], do: false, else: true
```
Rados13, line 159:
```suggestion
    :ets.lookup(@rooms_to_tabels, room_id) != []
```

lib/jellyfish/component/hls/request_handler.ex
```diff
@@ -0,0 +1,216 @@
...
+  end
+
+  defp is_partial_ready?(_partial, nil) do
+    false
+  end
+
+  defp is_partial_ready?(partial, last_partial),
+    do: partial_to_integer(last_partial) >= partial_to_integer(partial)
+
+  defp partial_to_integer({segment_sn, partial_sn}), do: segment_sn * 100 + partial_sn
```
Rados13, line 215: Why 😭 ?

Karolk99: I forgot about it

lib/jellyfish/component/hls/request_handler.ex
```diff
@@ -0,0 +1,208 @@
...
+  def handle_file_request(room_id, filename) do
+    path = Path.join([@hls_directory, room_id, filename])
+    File.read(path)
+  end
+
+  @doc """
+  Handles ll-hls partial requests
+  """
+  @spec handle_partial_request(Room.id(), String.t(), non_neg_integer()) ::
+          {:ok, binary()} | {:error | atom()}
```
sgfn, line 48:
```suggestion
          {:ok, binary()} | {:error, atom()}
```

lib/jellyfish/component/hls/request_handler.ex
```diff
@@ -0,0 +1,208 @@
...
+    unless is_partial_ready(partial, last_partial) do
+      wait_for_manifest_ready(room_id, partial, :manifest)
+    end
+
+    EtsHelper.get_manifest(room_id)
+  end
+
+  @doc """
+  Handles manifest requests with specific partial requested (ll-hls)
+  """
```
sgfn, line 69: Same doc as the function above

lib/jellyfish/component/hls/request_handler.ex
```diff
@@ -0,0 +1,208 @@
...
+  @spec handle_partial_request(Room.id(), String.t(), non_neg_integer()) ::
+          {:ok, binary()} | {:error | atom()}
+  def handle_partial_request(room_id, filename, offset) do
+    EtsHelper.get_partial(room_id, filename, offset)
+  end
+
+  @doc """
+  Handles manifest requests with specific partial requested (ll-hls)
+  """
+  @spec handle_manifest_request(Room.id(), partial()) :: {:ok, binary()} | {:error, any()}
```
sgfn, line 56: For consistency
```suggestion
  @spec handle_manifest_request(Room.id(), partial()) :: {:ok, String.t()} | {:error, any()}
```

lib/jellyfish/component/hls/request_handler.ex
```diff
@@ -0,0 +1,208 @@
...
+    {:stop, :normal, state}
+  end
+
+  @impl true
+  def terminate(_reason, %{room_id: room_id}) do
+    EtsHelper.remove_room(room_id)
+  end
+
+  @impl true
+  def handle_info({:DOWN, _ref, :process, pid, _reason}, %{room_pid: pid} = state) do
```
sgfn, line 142: Shouldn't `pid` be pinned here? Or used in a `when` clause?

lib/jellyfish/component/hls/request_handler.ex
```diff
@@ -0,0 +1,208 @@
...
+        is_same_segment_greater_or_equal_partial_sn(partial, last_partial)
+
+  defp is_last_segment_sn_greater({segment_sn, _partial_sn}, {last_segment_sn, _last_partial_sn}),
+    do: last_segment_sn > segment_sn
+
+  defp is_same_segment_greater_or_equal_partial_sn(
+         {segment_sn, partial_sn},
+         {last_segment_sn, last_partial_sn}
+       ),
+       do: last_segment_sn == segment_sn and last_partial_sn >= partial_sn
```
sgfn, line 207: How about something like this?
```suggestion
  defp is_partial_ready({segment_sn, partial_sn}, {last_segment_sn, last_partial_sn}) do
    cond do
      last_segment_sn > segment_sn -> true
      last_segment_sn < segment_sn -> false
      true -> last_partial_sn >= partial_sn
    end
  end
```

lib/jellyfish/component/hls/ll_storage.ex
```diff
@@ -0,0 +1,184 @@
...
+    unless filename == "index.m3u8" do
+      add_manifest_to_ets(filename, content, state)
+      send_update(filename, state)
+    end
+
+    {result, state}
+  end
+
+  defp add_manifest_to_ets(filename, manifest, %{table: table}) do
+    if String.contains?(filename, "_delta.m3u8") do
```
sgfn, line 108:
```suggestion
    if String.ends_with?(filename, "_delta.m3u8") do
```

sgfn (changes requested):
Nicely done :rocket: Small issues and nitpicks

Necessary in the full PR:
- API spec update
- Docs update
- HLS controller test
- UT for new modules in `hls/`, where possible

lib/jellyfish/component/hls/request_handler.ex
Karolk99: it checks whether they are the same or not

lib/jellyfish_web/controllers/hls_controller.ex
```diff
@@ -0,0 +1,81 @@
+defmodule JellyfishWeb.HLSController do
+  use JellyfishWeb, :controller
+
+  require Logger
+  alias Jellyfish.Component.HLS.RequestHandler
+  alias Plug.Conn
+
+  @hls_directory "jellyfish_output/hls_output"
```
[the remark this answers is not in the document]
Karolk99: I don't want to call it each time there is a request. So I will add it to the config.

Rados13 (commented):
I also agree with the issues made by @sgfn (very good code review).

test/jellyfish/component/hls/ll_storage_test.exs
```diff
@@ -0,0 +1,162 @@
...
+    {:ok, @manifest_content} = EtsHelper.get_manifest(room_id)
+    {:ok, @partial_sn} = EtsHelper.get_recent_partial(room_id)
+
+    manifest_path = Path.join(directory, @manifest_name)
+    {:ok, @manifest_content} = File.read(manifest_path)
+
+    pid = self()
+
+    spawn(fn ->
+      RequestHandler.handle_manifest_request(room_id, @partial_sn)
```
sgfn, line 74:
```suggestion
      {:ok, @manifest_content} = RequestHandler.handle_manifest_request(room_id, @partial_sn)
```
(not sure if I entered the correct constant)

sgfn (changes requested):
Very nice tests 🥇 

What I'm gonna need before the go-ahead:
1. Update `lib/jellyfish_web/api_spec/`
2. Update docs
3. Before merging, run `mix api.spec` and `./compile_proto.sh`

I'll insist on refactoring `RequestHandler.is_partial_ready`: if you don't like my suggestion, at least rename the 43-char long function to something shorter :smile:

test/jellyfish_web/controllers/hls_controller_test.exs
```diff
@@ -0,0 +1,165 @@
...
+
+    on_exit(fn -> :file.del_dir_r(output_path) end)
+  end
+
+  test "request master manifest", %{conn: conn} do
+    conn1 = get(conn, ~p"/hls/#{@room_id}/#{@master_manifest_name}")
+    assert @master_manifest_content == response(conn1, 200)
+
+    conn2 = get(conn, ~p"/hls/#{@wrong_room_id}/#{@master_manifest_name}")
+    response(conn2, 404)
```
Rados13, line 45: Don't we send any data in response e.g: description?

Rados13 (changes requested):

lib/jellyfish/component/hls/ets_helper.ex
```diff
@@ -0,0 +1,148 @@
...
+      :ets.insert(@rooms_to_tables, {room_id, table})
+      {:ok, table}
+    end
+  end
+
+  @spec remove_room(Room.id()) :: :ok | {:error, String.t()}
+  def remove_room(room_id) do
+    case :ets.lookup(@rooms_to_tables, room_id) do
+      [{^room_id, table}] ->
+        if table_exists?(table), do: :ets.delete(table)
```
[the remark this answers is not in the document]
Karolk99: RequestHandler calls `remove_room` function but the process responsible for creating ets table is the storage process. When storage dies before RequestHandler call `remove_room` ets will be already removed.

test/jellyfish/component/hls/ll_storage_test.exs
```diff
@@ -0,0 +1,162 @@
...
+
+    %{storage: storage, directory: directory, room_id: room_id}
+  end
+
+  test "store segment", %{storage: storage, directory: directory} do
+    {:ok, _storage} = store_segment(storage)
+
+    segment_path = Path.join(directory, @segment_name)
+    {:error, :enoent} = File.read(segment_path)
+  end
```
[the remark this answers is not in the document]
Karolk99: in case of ll-hls we don't want to save whole segment because we're already saving partials that at the end represents whole segment

test/jellyfish/component/hls/ll_storage_test.exs
```diff
@@ -0,0 +1,162 @@
...
+    {:ok, @delta_manifest_content} = File.read(manifest_path)
+
+    pid = self()
+
+    spawn(fn ->
+      RequestHandler.handle_delta_manifest_request(room_id, @partial_sn)
+      send(pid, :manifest)
+    end)
+
+    assert_receive(:manifest)
```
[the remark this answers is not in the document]
Karolk99: i change it to test timeout

test/jellyfish/component/hls/request_handler_test.exs
```diff
@@ -0,0 +1,148 @@
...
+    {:ok, @manifest_content} = RequestHandler.handle_manifest_request(room_id, @partial)
+
+    pid = self()
+
+    spawn(fn ->
+      {:ok, @manifest_content} = RequestHandler.handle_manifest_request(room_id, @next_partial)
+      send(pid, :manifest)
+    end)
+
+    refute_receive(:manifest, 500)
```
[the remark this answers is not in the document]
Karolk99: I can't change it like in the previous example because handle_manifest_request is blocking

sgfn (approved):
LGTM, don't forget to write about the changes in the docs :pray:

test/jellyfish/component/hls/ets_helper_test.exs
```diff
@@ -32,61 +32,64 @@ defmodule Jellyfish.Component.HLS.EtsHelperTest do
 
   test "rooms managment" do
     room_id = UUID.uuid4()
-    {:error, :room_not_found} = EtsHelper.get_partial(room_id, @partial_name, @offset)
+    assert {:error, :room_not_found} == EtsHelper.get_partial(room_id, @partial_name, @offset)
```
Rados13, line 35: [Assert works with pattern match](https://hexdocs.pm/ex_unit/1.12.3/ExUnit.Assertions.html#assert/1), so changing `=` to `==` wasn't necessary.

test/jellyfish/component/hls/request_handler_test.exs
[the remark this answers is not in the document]
Rados13: I think in this case you should use Task instead of spawn, and instead of this strange sending message and refute receive, use [Task.yield](https://hexdocs.pm/elixir/1.13/Task.html#yield/2).

Rados13 (approved):
Only one last change previous issues

test/jellyfish/component/hls/request_handler_test.exs
```diff
@@ -0,0 +1,148 @@
...
+
+    RequestHandler.update_recent_partial(room_id, @next_partial)
+
+    assert_receive(:manifest, 1000)
```
Rados13, line 82:
```suggestion
    task=Task.async(fn ->
      RequestHandler.handle_manifest_request(room_id, @next_partial)
    end)
    assert nil = Task.yield(task)
    
    RequestHandler.update_recent_partial(room_id, @next_partial)

    Task.await(task, 1000)
    ```

PR #68, opened 2023-08-10
[RTC-308] Make simulcast configurable

sgfn:
ref:
- https://github.com/jellyfish-dev/elixir_server_sdk/pull/27
- https://github.com/jellyfish-dev/jellyfish-docs/pull/18

### DIFF
--- a/lib/jellyfish/room.ex
+++ b/lib/jellyfish/room.ex
@@ -73,9 +73,10 @@ defmodule Jellyfish.Room do
     end
   end
 
-  @spec add_peer(id(), Peer.peer()) :: {:ok, Peer.t()} | {:error, :reached_peers_limit}
-  def add_peer(room_id, peer_type) do
-    GenServer.call(registry_id(room_id), {:add_peer, peer_type})
+  @spec add_peer(id(), Peer.peer(), map() | nil) ::
+          {:ok, Peer.t()} | :error | {:error, :reached_peers_limit}
+  def add_peer(room_id, peer_type, options) do
+    GenServer.call(registry_id(room_id), {:add_peer, peer_type, options})
   end
 
   @spec set_peer_connected(id(), Peer.id()) ::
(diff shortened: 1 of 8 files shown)


lib/jellyfish/room.ex
```diff
@@ -72,9 +72,10 @@ defmodule Jellyfish.Room do
     end
   end
 
-  @spec add_peer(id(), Peer.peer()) :: {:ok, Peer.t()} | {:error, :reached_peers_limit}
-  def add_peer(room_id, peer_type) do
-    GenServer.call(registry_id(room_id), {:add_peer, peer_type})
+  @spec add_peer(id(), Peer.peer(), map() | nil) ::
```
roznawsk, line 75: I am wondering, whether the `options` shouldn't have a defined type instead of just `map`.
I'm not saying that we necessarily need this new type tho

roznawsk (approved):

lib/jellyfish/room.ex
```diff
@@ -72,9 +72,10 @@ defmodule Jellyfish.Room do
...
+          {:ok, Peer.t()} | :error | {:error, :reached_peers_limit}
+  def add_peer(room_id, peer_type, options) do
```
roznawsk, line 77:
```suggestion
  def add_peer(room_id, peer_type, options \\ %{}) do
```

lib/jellyfish/room.ex
```diff
@@ -125,23 +126,32 @@ defmodule Jellyfish.Room do
...
-
-        {{:ok, peer}, state}
+        options =
+          Map.merge(
+            %{
+              engine_pid: state.engine_pid,
+              network_options: state.network_options,
+              video_codec: state.config.video_codec
+            },
+            if(is_nil(options), do: %{}, else: options)
```
roznawsk, line 141:
```suggestion
            options
```

roznawsk (approved):

lib/jellyfish/room.ex
sgfn: Could be, but I think I'll leave it as-is ¯\_(ツ)_/¯

PR #72, opened 2023-08-23
[RTC-333] Limit HLS components to 1 per room

sgfn:


lib/jellyfish/room.ex
```diff
@@ -386,7 +391,21 @@ defmodule Jellyfish.Room do
...
+      video_codec != :h264 ->
+        {:error, :incompatible_codec}
+
+      Map.values(components) |> Enum.any?(&(&1.type == Component.HLS)) ->
+        {:error, :reached_components_limit}
+
+      true ->
+        :ok
+    end
+  end
```
Karolk99, line 408: What about:
```suggestion
  defp check_component_allowed(Component.HLS, %{config: %{video_codec: video_codec}, components: components}) do

    hls_components_present = components_includes_HLS?(components)

    case {video_codec, hls_components_present} do
      {:h264, false} -> :ok
      {:h264, true} -> {:error, :reached_components_limit}
      _ -> {:error, :incompatible_codec}
    end
  end

  defp components_includes_HLS?(components) do
    components
    |> Map.values()
    |> Enum.any?(&(&1.type == Component.HLS))
  end
```

Karolk99 (approved):

PR #81, opened 2023-09-05
Unique named partials

Karolk99:


lib/jellyfish/component/hls/request_handler.ex
```diff
@@ -162,13 +188,24 @@ defmodule Jellyfish.Component.HLS.RequestHandler do
...
+  defp wait_for_partial_ready(room_id, filename) do
+    GenServer.cast(registry_id(room_id), {:preload_hint, room_id, filename, self()})
+
+    receive do
+      :preload_hint_ready ->
+        EtsHelper.get_partial(room_id, filename)
+    end
+  end
+
+  defp update_status(status, last_partial, preload_hints) do
```
sgfn, line 200: I'd argue for renaming this function, as it currently has a side-effect of notifying pids that are waiting for partial/preload hint. Maybe something like `update_status_and_notify_waiting_pids` but shorter? Maybe even extract the notifying to another function?

lib/jellyfish/component/hls/request_handler.ex
```diff
@@ -162,13 +188,24 @@ defmodule Jellyfish.Component.HLS.RequestHandler do
...
     {waiting_pids, status} =
       status
       |> Map.put(:last_partial, last_partial)
       |> pop_in([:waiting_pids, last_partial])
 
     send_partial_ready(waiting_pids)
+    send_preload_hint_ready(preload_hints)
+
```
sgfn, line 208: Either I am missing something, or the `:preload_hints` list in state is never emptied after sending the notifications...

lib/jellyfish/component/hls/request_handler.ex
```diff
@@ -162,13 +188,24 @@ defmodule Jellyfish.Component.HLS.RequestHandler do
...
 
-  defp update_status(status, last_partial) do
+  defp wait_for_partial_ready(room_id, filename) do
+    GenServer.cast(registry_id(room_id), {:preload_hint, room_id, filename, self()})
+
+    receive do
+      :preload_hint_ready ->
+        EtsHelper.get_partial(room_id, filename)
+    end
+  end
```
sgfn, line 198: `wait_for_manifest_ready` and `wait_for_partial_ready`: two functions with similar names, yet totally different APIs. I'd say we should either rename one of them, or make them behave similarly

test/jellyfish/component/hls/ll_storage_test.exs
```diff
@@ -47,10 +47,10 @@ defmodule Jellyfish.Component.HLS.LLStorageTest do
   test "store partial", %{storage: storage, directory: directory, room_id: room_id} do
     {:ok, _storage} = store_partial(storage)
 
-    partial_path = Path.join(directory, @partial_name)
+    partial_path = Path.join(directory, @segment_name)
```
sgfn, line 50: I think some comment about the logic here would be appreciated

sgfn (changes requested):
Would it be possible to include some tests for the preload hints?

lib/jellyfish/component/hls/request_handler.ex
Karolk99: I don't see the problem those are private functions. The names describe really well what they do.

lib/jellyfish/component/hls/request_handler.ex
```diff
@@ -186,6 +223,34 @@ defmodule Jellyfish.Component.HLS.RequestHandler do
...
+  defp check_if_preload_hint(_partial_sn, _recent_partial_sn), do: false
+
+  # Filename example: muxed_segment_32_g2QABXZpZGVv_5_part.m4s
+  defp get_partial_sn(filename) do
+    filename
+    |> String.split("_")
+    |> Enum.filter(fn s -> match?({_integer, ""}, Integer.parse(s)) end)
+    |> Enum.map(fn sn -> String.to_integer(sn) end)
+    |> List.to_tuple()
+  end
```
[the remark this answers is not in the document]
Karolk99: If there will be any mistake it will return error file_not_found, so we will catch that. But I will try to fix this.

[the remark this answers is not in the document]
Karolk99: This solution is strictly for jellyfish so don't think we should be worried about different naming functions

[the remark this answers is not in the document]
sgfn: OK, fair enough

sgfn (approved):
💪

PR #82, opened 2023-09-07
Openapi generator compatibility

roznawsk:


LVala (approved):

lib/jellyfish_web/controllers/component_controller.ex
```diff
@@ -59,7 +61,7 @@ defmodule JellyfishWeb.ComponentController do
     ]
 
   def create(conn, %{"room_id" => room_id} = params) do
-    with component_options <- Map.get(params, "options", %{}),
+    with component_options <- Map.get(params, "options", nil),
```
sgfn, line 64: If we're changing this, I'd say we should change peer_controller to behave similarly

sgfn (approved):
https://previews.123rf.com/images/tigatelu/tigatelu1508/tigatelu150800030/44232896-szcz%C4%99%C5%9Bliwy-emotikon-bu%C5%BAk%C4%99-daj%C4%85c-kciuki-do-g%C3%B3ry.jpg

PR #83, opened 2023-09-11
[RTC-344] toggleable ll-hls

sgfn:
ref: https://github.com/jellyfish-dev/elixir_server_sdk/pull/30 https://github.com/jellyfish-dev/jellyfish-docs/pull/30

### DIFF
--- a/lib/jellyfish/component/hls.ex
+++ b/lib/jellyfish/component/hls.ex
@@ -8,7 +8,7 @@ defmodule Jellyfish.Component.HLS do
   alias Jellyfish.Component.HLS.{LLStorage, RequestHandler, Storage}
   alias Jellyfish.Room
 
-  alias JellyfishWeb.ApiSpec
+  alias JellyfishWeb.ApiSpec.Component.HLS.Options
 
   alias Membrane.RTC.Engine.Endpoint.HLS
   alias Membrane.RTC.Engine.Endpoint.HLS.{CompositorConfig, HLSConfig, MixerConfig}
@@ -24,7 +24,7 @@ defmodule Jellyfish.Component.HLS do
 
   @impl true
   def config(options) do
-    with {:ok, valid_opts} <- OpenApiSpex.cast_value(options, ApiSpec.Component.HLS.schema()) do
+    with {:ok, valid_opts} <- OpenApiSpex.cast_value(options, Options.schema()) do
       low_latency? = valid_opts.lowLatency
       hls_config = create_hls_config(options.room_id, low_latency?: low_latency?)
 
--- a/lib/jellyfish/component/hls.ex
+++ b/lib/jellyfish/component/hls.ex
@@ -6,8 +6,10 @@ defmodule Jellyfish.Component.HLS do
   @behaviour Jellyfish.Endpoint.Config
   @behaviour Jellyfish.Component
 
-  alias Jellyfish.Component.HLS.{LLStorage, RequestHandler}
+  alias Jellyfish.Component.HLS.{LLStorage, RequestHandler, Storage}
   alias Jellyfish.Room
+
+  alias JellyfishWeb.ApiSpec
 
   alias Membrane.RTC.Engine.Endpoint.HLS
   alias Membrane.RTC.Engine.Endpoint.HLS.{CompositorConfig, HLSConfig, MixerConfig}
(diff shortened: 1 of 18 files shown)


lib/jellyfish/component/hls.ex
```diff
@@ -20,34 +22,37 @@ defmodule Jellyfish.Component.HLS do
...
+               aligned: true
+             }
            }
+         },
+         hls_config: %HLSConfig{
+           hls_mode: :muxed_av,
+           mode: :live,
+           target_window_duration: :infinity,
+           segment_duration: @segment_duration,
+           partial_segment_duration: @partial_segment_duration,
```
Karolk99, line 49: right now in both scenarios, SinkBin will create partial segments.
```suggestion
           partial_segment_duration: (if validopts.lowLatency, do: @partial_segment_duration, else: nil),
```

lib/jellyfish_web/controllers/peer_controller.ex
```diff
@@ -67,7 +67,7 @@ defmodule JellyfishWeb.PeerController do
     # pid and adding a new peer to it
     # in such a case, the controller will fail
     # and Phoenix will return 500
-    with peer_options <- Map.get(params, "options", %{}),
```
Karolk99, line 70: Maybe I missed something but this change just adds one additional `if` statement to the `room.ex` file.

Karolk99 (changes requested):
We should add a test in `component_controller_test.ex`.
Also, I thought about a `lowLatency` field in the HLS metadata. Right now server won't know if the HLS stream is low latency or not.

lib/jellyfish/component/hls.ex
sgfn: 😵‍💫

lib/jellyfish/component/hls.ex
```diff
@@ -20,34 +22,37 @@ defmodule Jellyfish.Component.HLS do
...
-       output_directory: output_dir(options.room_id),
-       mixer_config: %MixerConfig{
-         video: %CompositorConfig{
-           stream_format: %Membrane.RawVideo{
-             width: 1920,
-             height: 1080,
-             pixel_format: :I420,
-             framerate: {24, 1},
-             aligned: true
+    with {:ok, valid_opts} <- OpenApiSpex.cast_value(options, ApiSpec.Component.HLS.schema()) do
```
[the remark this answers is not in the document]
sgfn: That's because RTSP is further working on the options passed to `cast_value`, meanwhile we're only using one flag here in HLS. Same thing happens in WebRTC with `enableSimulcast` flag

lib/jellyfish_web/controllers/peer_controller.ex
sgfn: That's for consistency with changes from component_controller.ex in https://github.com/jellyfish-dev/jellyfish/pull/82

sgfn:
Don't review just yet, we need to talk about the `metadata` fields

lib/jellyfish_web/controllers/peer_controller.ex
roznawsk: I think we should go back to `Map.get(params, "options", %{})`, both when fetching peer options as well as component options.
My previous change was motivated by the fact, that the `options` could be `nil`, so a default value of `nil` made sense.
Now, neither peer nor component options are nullable, so the default value can be an empty map.

sgfn: Changed

lib/jellyfish_web/api_spec/component.ex
```diff
@@ -29,36 +28,26 @@ defmodule JellyfishWeb.ApiSpec.Component do
...
+      HLS,
+      RTSP
+    ],
+    discriminator: %OpenApiSpex.Discriminator{
+      propertyName: "type",
+      mapping: %{
+        "hls" => HLS,
+        "rtsp" => RTSP
+      }
+    }
```
sgfn, line 51: Without the discriminator, the `component_controller` tests fail at `assert_response_schema` due to "multiple schemas match". That's because there are no fields in RTSP component metadata, and an empty map matches every other map.

lib/jellyfish_web/api_spec/component/rtsp.ex
```diff
@@ -4,43 +4,75 @@ defmodule JellyfishWeb.ApiSpec.Component.RTSP do
...
+  defmodule Metadata do
+    @moduledoc false
+
+    require OpenApiSpex
+
+    OpenApiSpex.schema(%{
+      title: "ComponentMetadataRTSP",
+      description: "Metadata specific to the RTSP component",
+      type: :object,
+      properties: %{}
```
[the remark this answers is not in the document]
roznawsk: It probably won't. However since this is a valid API spec and it seems that quite often we have a case when we want to have an empty schema, maybe we should fix it in the python generator.

roznawsk (approved):

lib/jellyfish_web/api_spec/component/rtsp.ex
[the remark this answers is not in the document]
sgfn: Update: we have verified that (thanks to the discriminator) the Python generator has no issues with this, hooray

test/jellyfish_web/controllers/component_controller_test.exs
```diff
@@ -67,6 +73,37 @@ defmodule JellyfishWeb.ComponentControllerTest do
       assert response(room_conn, :no_content)
     end
 
+    test "renders component with ll-hls enabled", %{conn: conn} do
```
Karolk99, line 76: I would add a check for the request handler whether it's spawned or not

Karolk99 (approved):
LGTM (let's get this merged)

PR #84, opened 2023-09-13
add creating multipart room

blazpie:
This PR adds the ability to create a multipart room on jellyfishes in the cluster.
Whole communication between rooms is made via Phoenix.Pubsub.
After creating room a room's process subscribes to PubSub topic with it's id, and broadcasts `:room_created` to other rooms with same id.
When receives `:room_created` room makes direct broadcast to room which creates new part of room with `:room_exists`
When room receives `:room_exists` it adds remote endpoint to the RTC engine and sends back to other room link proposal created by remote endpoint.
When room receives link proposal it creates another remote endpoint with data from link proposal.

Further communication is made between Remote Endpoints


lib/jellyfish/room.ex
```diff
@@ -117,6 +119,8 @@ defmodule Jellyfish.Room do
   @impl true
   def init([id, max_peers, video_codec]) do
     state = new(id, max_peers, video_codec)
+    Phoenix.PubSub.subscribe(Jellyfish.PubSub, id)
+    Phoenix.PubSub.broadcast(Jellyfish.PubSub, id, {:room_created, Node.self()})
```
Rados13, line 123: I think we should check whether broadcasting succeeds.
```suggestion
    :ok = Phoenix.PubSub.broadcast(Jellyfish.PubSub, id, {:room_created, Node.self()})
```

lib/jellyfish/room.ex
```diff
@@ -295,6 +299,79 @@ defmodule Jellyfish.Room do
...
+        Phoenix.PubSub.direct_broadcast(
+          node,
+          Jellyfish.PubSub,
+          room_id,
+          {:create_remote_endpoint, link_proposal, Node.self()}
+        )
+    end
+
+    {:noreply, state}
+  end
```
Rados13, line 357: The whole logic could be simplified. When a user provides the id of the room, we know that he wants to create a distributed. In this situation, we add a remote endpoint and broadcast message to other rooms from this endpoint that we start distributed jellyroom. Other rooms receive a message about starting a distributed room, and they add remoteEndpoint if it doesn't already exist. The biggest drawback of the current implementation is that we have to maintain information about nodes and in a situation of node crash we have to make it consistent.

lib/jellyfish/room_service.ex
```diff
@@ -55,27 +55,39 @@ defmodule Jellyfish.RoomService do
...
+  #   {min_node, _room_size} =
+  #     Enum.min_by(node_resources, fn {_node_name, room_num} -> room_num end)
+
+  #   if Enum.count(node_resources) > 1 do
+  #     Logger.info("Node with least used resources is #{inspect(min_node)}")
+  #     GenServer.call({__MODULE__, min_node}, {:create_room, max_peers, video_codec, id})
+  #   else
+  #     GenServer.call(__MODULE__, {:create_room, max_peers, video_codec, id})
+  #   end
+  # end
```
Rados13, line 87: It can be very easily solved if room_id is different than nil we start It on a current node in all other situations we use load balancing.

Rados13 (changes requested):
We should also add some tests for it.

lib/jellyfish/room.ex
blazpie: That logic also causes troubles:

1. We actually don't know if a user is creating a distributed room. API allows to pass ID but it's not checking if the room with this ID exists on other nodes (it's OK, it shouldn't be API's responsibility). If we decide to check it somewhere it'll be the same logic, but moved. 

2. In this implementation I decided to create Remote Endpoints as pairs. Creating RE without info that a room exists somewhere else makes no sense because it'll cause us to implement new logic when the room is distributed to more than two nodes.

IMO it's better to create an RE in response to info that a room exists, even if we need to handle node crash. I'm not sure if need to keep information about nodes. I left it in room's state, but the most actual information about linked nodes is in the RE state. Also RE has a logic to update partner node address by sending a handshake message at any time. Maybe we should remove nodes info from room state as it'll probably be useless after linking - all further messages will be send via RE

lib/jellyfish/room_service.ex
blazpie: That was my first idea but IMO that's not a good assumption. API should be clear and consequent. Passing or not room id value should not change the logic of how  and where room is created. We can't leave such a surprise in API. If we want to do it we should add a flag like `distributed: true` to make it explicit.

lib/jellyfish/room.ex
Rados13: 1. When the client wants to create a second engine on a different node, he always has to pass ID while creating a room on an other node. As for passing ID, while creating the first engine, I don't see much problem because the only cost we will get is to spawn an additional endpoint that will dump every received message, so this is not a problem. Also, the whole beauty of this solution is that we don't have to check if the room exists on the other node. We only inform other nodes that we are ready for this room to work distributed.
2. I don't understand this point, why we would need different logic.

Yeah I agree about that jellyfish should have minimal information about nodes and if anywhere it should be moved to RE.

lib/jellyfish/room_service.ex
Rados13: From one point it's true but, from the second point of view passing ID when you create room on the first jellyfish is useless. When you pass ID and want to use load balancing there is a question of how it should be handled when some room with this ID already exists etc.

mickel8:
Unpining myself for now

mickel8:
I am closing as we changed our priorities

PR #85, opened 2023-09-13
Fix get resource usage

Rados13:


lib/jellyfish/room.ex
```diff
@@ -278,8 +278,16 @@ defmodule Jellyfish.Room do
...
+    pid = state.engine_pid
+
+    try do
+      forwarded_tracks = Engine.get_num_forwarded_tracks(pid)
+      {:reply, forwarded_tracks, state}
+    catch
+      :exit, {:noproc, {GenServer, :call, [^pid, :get_num_forwarded_tracks, _timeout]}} ->
+        Logger.warning("Call to room #{state.id} failed")
+        {:reply, 0, state}
+    end
```
mickel8, line 290: I wonder how we should handle this. We call `get_num_forwarded_tracks` in RoomService in Task.async. If Engine crashesh, our room should exit anyway so the question is: should we crash room we engine crashesh or receive `DOWN` message from the Engine and close room in a more controlled way?

Rados13: I think we should stop the room process when we receive a `DOWN` message, but here we must also catch an error to not fail a `Task.async`.

lib/jellyfish/room_service.ex
```diff
@@ -101,7 +101,9 @@ defmodule Jellyfish.RoomService do
 
     room_ids
     |> Enum.map(fn room_id ->
-      Task.async(fn -> Room.get_num_forwarded_tracks(room_id) end)
+      Task.Supervisor.async(Jellyfish.TaskSupervisor, fn ->
```
mickel8, line 104: Task.Supervisor.async_nolink + Task.yield + Task.shutdown?

test/jellyfish_web/controllers/room_controller_test.exs
```diff
@@ -144,6 +144,25 @@ defmodule JellyfishWeb.RoomControllerTest do
       assert Enum.any?(rooms, &(&1.id == room2_id))
       assert Enum.all?(rooms, &(&1.id != room_id))
     end
+
+    test "roomService removes room on engine crash", %{room_id: room_id} = state do
```
mickel8, line 148: I would call this `room closes engine crash` as RoomService doesn't remove room in this test case

test/jellyfish_web/controllers/room_controller_test.exs
```diff
@@ -144,6 +144,25 @@ defmodule JellyfishWeb.RoomControllerTest do
...
+
+      room_pid = RoomService.find_room!(room_id)
+
+      :erlang.trace(Process.whereis(RoomService), true, [:receive])
+
+      %{engine_pid: engine_pid} = :sys.get_state(room_pid)
+
+      assert true = Process.exit(engine_pid, :error)
+
+      assert_receive({:trace, _pid, :receive, {:DOWN, _ref, :process, ^room_pid, :error}})
```
mickel8, line 159: We don't need this and `:erlang.trace` too? It should be enought to make sure `list_rooms` doesnt return room with the id `room_id`

lib/jellyfish/room_service.ex
```diff
@@ -176,7 +178,7 @@ defmodule Jellyfish.RoomService do
   def handle_info({:DOWN, _ref, :process, pid, reason}, state) do
     {room_id, state} = pop_in(state, [:rooms, pid])
 
-    Logger.warning("Process #{room_id} is down with reason: #{reason}")
+    Logger.warning("Process #{room_id} is down with reason: #{inspect(reason)}")
```
mickel8, line 181: If reason is an atom or string we dont need inspect here

mickel8 (changes requested):

lib/jellyfish/room_service.ex
Rados13: Yes I added this because once I receive something different, if I remember it was a tuple and this leads to error but I can remove it.

test/jellyfish_web/controllers/room_controller_test.exs
Rados13: Nope, because without it, `list_roms` returns room with id `room_id`. I think that because a room isn't removed from the registry fast enough and from that we get race condition.

roznawsk (approved):
LGTM

mickel8 (approved):

PR #86, opened 2023-09-15
Add grafana dashboard

Rados13:


Davenury (commented):
That's it for now. When I see the definitions and types of metrics, I'll probably be able to come up with some more ideas :smile:

mickel8:
Looks cool :+1:  I would just remove RSS and logo for now and take care of plot colors 

Thanks @Davenury for the CR :slightly_smiling_face:

mickel8 (approved):

grafana/jellyfish-server-overiew.json
```diff
@@ -0,0 +1,758 @@
...
+                    "url": "https://jellyfish-dev.github.io/jellyfish-docs/"
+                }
+            ],
+            "options": {
+                "code": {
+                    "language": "plaintext",
+                    "showLineNumbers": false,
+                    "showMiniMap": false
+                },
+                "content": "This is Jellyfish dashboard\n\n![Text](https://avatars.githubusercontent.com/u/119447297?s=400&u=f714adece660ae05d4b177d1d4e7d7886aa4e9d5&v=4)",
```
Davenury, line 51: Based on the last commit message seems like you wanted to delete those panels :smile: Seems like you forgot to save before exporting dashboard

mickel8:
Yeah exactly! But because we are on early stages of the project development, I think we don't need them right now. Especially that the project name and logo are subjects to change :)

PR #90, opened 2023-09-20
[RTC-345] Allow to configure HTTP endpoint IP

mickel8:
We should by default bind to the 0.0.0.0 or 0.0.0.0.0.0.0.0 only when running in docker. When we run behind proxy it's better to run on a loopback not to accidentaly expose JF to the outside. 

We also allow for configuring this IP env so that everyone can run on whatever port they wish

### DIFF
--- a/config/prod.exs
+++ b/config/prod.exs
@@ -9,7 +9,7 @@ config :logger, level: :info
 
 # run the server automatically when using prod release
 config :jellyfish, JellyfishWeb.Endpoint,
-  http: [ip: {0, 0, 0, 0, 0, 0, 0, 0}, port: 8080],
+  http: [ip: {127, 0, 0, 1}, port: 8080],
   server: true
 
 # Runtime production configuration, including reading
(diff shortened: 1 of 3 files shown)


sgfn (approved):

roznawsk (approved):

config/prod.exs
```diff
@@ -9,7 +9,7 @@ config :logger, level: :info
 
 # run the server automatically when using prod release
 config :jellyfish, JellyfishWeb.Endpoint,
-  http: [ip: {0, 0, 0, 0, 0, 0, 0, 0}, port: 8080],
+  http: [ip: {127, 0, 0, 1}, port: 8080],
```
Rados13, line 12: Why do we change it? Most of the time, we run prod in docker, so changing it to 127.0.0.1 doesn't make much sense.

Rados13 (commented):
What is the sense of all of these changes? Becas

config/prod.exs
mickel8: For docker we override this variable to 0.0.0.0 and because we run Jellyfish behind nginx (there is no way to configure SSL directly in Jellyfish right now), I think it is safer not to expose Jellyfish to the outside by default

Rados13 (approved):

PR #91, opened 2023-09-20
[RTC-345] Prefix env vars with JF

mickel8:
Let's follow good practices and perfix our env vars with `JF` not to overlap with other software. 

I also deleted `JELLYFISH_ADDRESS`. The idea is that `JF_HOST` is how we are seen by the others and it's used both for generating URLs and connecting to correct Jellyfish (i.e. address returned when we create a room).

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -1,19 +1,76 @@
-# used by the server e.g. to create tokens
-SECRET_KEY_BASE=super-secret-key
+# JF_IP and JF_PORT defines ip and port
+# our HTTP endpoint socket will listen to.
+# The default for Docker builds is 0.0.0.0 to allow 
+# access from the outside of container.
+# Note that this cannot be 127.0.0.1 as 
+# in Docker it only allows for traffic
+# from within the container.
+# For other builds, it is 127.0.0.1 not to 
+# accidentaly expose Jellyfish to the outside world
+# when it runs behind some proxy like nginx.
+# 
+# JF_IP=0.0.0.0
+# JF_PORT=5002
 
-# true, if WebRTC peers are used
-WEBRTC_USED=true
 
-# hostname used to generate URLs throught the server
-VIRTUAL_HOST=localhost
-PORT=5002
-SERVER_API_TOKEN=development
+# JF_HOST and JF_HOST_PORT defines how Jellyfish
+# should be seen from the outside.
+# When you run Jellyfish behind proxy, your
+# Jellyfish will run on 127.0.0.1 or 0.0.0.0
+# (see JF_IP and JF_PORT for more information)
+# but you don't want to and in fact you can't 
+# use those addresses for generating URLs, or 
+# telling Jellyfish Client to which server instance it 
+# should connect to (when running Jellyfish in a cluster).
+# That's why we have separate environment variables.
+#
+# Example
+#
+# You run Jellyfish on a machine with some public 
+# ip address `PUB_IP` and domain `DOMAIN`.
+# You most likely want to provide the following 
+# configuration when running Jellyfish using Docker
+# 
+# JF_HOST=$DOMAIN or $PUB_IP
+# JF_HOST_PORT=443
+# JF_IP = 0.0.0.0
+# JF_PORT = 5002
+# 
+# JF_IP and JF_PORT are set by default
+# so everything you need to set is JF_HOST and JF_HOST_PORT
+# 
+JF_HOST=localhost
+JF_HOST_PORT=443
+
+# JF_METRICS_IP=0.0.0.0
+# JF_METRICS_PORT=9568
+
+# Token used for authorizing HTTP requests
+JF_SERVER_API_TOKEN=jellyfish_docker_token
+
+# Used by the server e.g. to create client tokens.
+# If not set, it will be generated
+# JF_SECRET_KEY_BASE=super-secret-key
 
 # Decide if jellyfish will check origin of requests
-# CHECK_ORIGIN=false
+# JF_CHECK_ORIGIN=true
+
+# Where Jellyfish should save its artifacts
+# You can get access to this directory e.g. by mounting 
+# a volume with:
+#
+#   -v $(pwd)/jellyfish_output:/app/jellyfish_output 
+#
+# JF_OUTPUT_BASE_PATH=/app/jellyfish_output
+
+
+# WEBRTC ENVS
+
+# true, if WebRTC peers are used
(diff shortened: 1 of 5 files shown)


sgfn (approved):
LGTM, don't forget to update the docs :pray:

config/runtime.exs
```diff
@@ -11,62 +11,62 @@ alias Jellyfish.ConfigReader
...
-  integrated_turn_tcp_port: ConfigReader.read_port("INTEGRATED_TURN_TCP_PORT"),
+    ConfigReader.read_port_range("JF_INTEGRATED_TURN_PORT_RANGE") || {50_000, 59_999},
+  integrated_turn_tcp_port: ConfigReader.read_port("JF_INTEGRATED_TURN_TCP_PORT"),
   jwt_max_age: 24 * 3600,
-  output_base_path: System.get_env("OUTPUT_BASE_PATH", "jellyfish_output") |> Path.expand(),
-  address: System.get_env("JELLYFISH_ADDRESS") || "#{host}:#{port}",
-  metrics_ip: ConfigReader.read_ip("METRICS_IP") || {127, 0, 0, 1},
-  metrics_port: ConfigReader.read_port("METRICS_PORT") || 9568
+  output_base_path: System.get_env("JF_OUTPUT_BASE_PATH", "jellyfish_output") |> Path.expand(),
+  address: "#{host}:#{port}",
```
Rados13, line 52: Why did you remove `JELLYFISH_ADDRESS` variable?

mickel8: As in the PR description :stuck_out_tongue:

Rados13: Are you sure that this won't break when you use for example proxy? Besides IMO it will look strange if somebody set `JF_HOST` to some domain and then jellyfish will send his clients to connect to `domain`:`PORT`.

mickel8: Why? #{host}:#{port} was default in the previous implementation. That's also our HTTP server address and when we open WS connection we can use FQDN, we don't have to use ip address. We also don't use this for WebRTC stuff so I would try with this setup as it limits the number of env vars.

mickel8: After discussing the problem offline, I added JF_HOST_PORT.

The problem was that without JF_HOST_PORT, we would return invalid port to Jellyfish Client .e.g 5002 instead of 443 (when running behind proxy) so we either have to allow to pass port in JF_HOST or add additional env var. I opt for additional env var as it is easier to understand and describe those 4 variables (JF_HOST, JF_HOST_PORT, JF_IP, JF_PORT). See .env.sample for more detailed explanation.

cc @sgfn @roznawsk

.env.sample
```diff
@@ -1,19 +1,76 @@
...
+#
+#   -v $(pwd)/jellyfish_output:/app/jellyfish_output 
+#
+# JF_OUTPUT_BASE_PATH=/app/jellyfish_output
+
+
+# WEBRTC ENVS
+
+# true, if WebRTC peers are used
+JF_WEBRTC_USED=true
```
Rados13, line 70: Why do some variable defaults are commented out and some not? Shouldn't we simply provide default values for all variables where it is possible?

config/runtime.exs
```diff
@@ -78,5 +85,5 @@ case System.get_env("SERVER_API_TOKEN") do
 end
 
 if prod? do
-  config :jellyfish, JellyfishWeb.Endpoint, url: [host: host, port: 443, scheme: "https"]
+  config :jellyfish, JellyfishWeb.Endpoint, url: [scheme: "https"]
```
Rados13, line 88: Could we move all setting up configs about `JellyfishWeb.Endpoint` closes together and the second group would configs under key `:jellyfish`? By this I mean moving this if after `check_origin` if and moving `JF_SERVER_API_TOKEN` before `config :jellyfish, JellyfishWeb.Endpoint`.

.env.sample
```diff
@@ -1,19 +1,76 @@
...
+# JF_IP and JF_PORT defines ip and port
+# our HTTP endpoint socket will listen to.
+# The default for Docker builds is 0.0.0.0 to allow 
+# access from the outside of container.
+# Note that this cannot be 127.0.0.1 as 
+# in Docker it only allows for traffic
+# from within the container.
+# For other builds, it is 127.0.0.1 not to 
+# accidentaly expose Jellyfish to the outside world
+# when it runs behind some proxy like nginx.
```
roznawsk, line 10:
```suggestion
# In Docker set it to 0.0.0.0 to allow 
# access from outside of the container.
# Note that this cannot be 127.0.0.1 as 
# in Docker it only allows for traffic
# from within the container.
# For other builds, set it to 127.0.0.1 to 
# avoid accidentally exposing Jellyfish to the outside world
# when it runs behind some proxy like nginx.
```

roznawsk (changes requested):

sgfn (commented):
Besides issues outlined by @roznawsk and @Rados13, LGTM

I agree that we should limit the comments in the `.env` files to an absolute minimum, 1-2 lines max per variable. Anything else we might need to tell the users about should be in the docs

.env.sample
mickel8: Uncommented are the ones needed to be set to get JF running locally

docker-compose.yaml
```diff
@@ -47,7 +47,8 @@ services:
       <<: *jellyfish-environment
       RELEASE_NODE: app@app2
       NODE_NAME: app@app2
-      PORT: 4002
+      JF_HOST: "localhost:4002"
```
mickel8, line 50: I am not sure if I am gonna introduce this change as IMO its misleading now :thinking:

.env.sample
Rados13: `JF_WEBRTC_USED` isn't needed because you could use JF without WebRTC, so if `JF_WEBRTC_USED` is set `JF_OUTPUT_BASE_PATH` should be set also I think.

Rados13 (approved):
Besides your doubts, LGTM.

roznawsk (approved):
👍🏽 

I think it would be best if you introduced those changes to elixir and python SDK and then merge everything at the same time

docker-compose.yaml
mickel8: If we had 4 vars we could get rid of `:4002` but let's try with 3 vars and see how it works

.env.sample
mickel8: Yeah but I aimed at WebRTC usage as it's most popular

PR #92, opened 2023-09-20
[RTC-345] Introduce JF dedicated env vars for running in a distributed mode

mickel8:
This PR introduces our own env variables for running JF in a cluster.

When we run erlang/elixir locally, we can setup distribution using flags like `--sname` `--name` `--cookie` etc. However, when we make a release, we are provided with special env variables e.g. `RELEASE_NODE` `RELEASE_COOKIE` etc. which results in multiple configuration options depending on environment we run Jellyfish in.

Intoducing our own env vars we:
* unify how JF is configured
* have better env names e.g. JF_COOKIE instead of strange RELEASE_COOKIE
* make **future** documentation easier to understand as our env vars are always available and always the same. RELEASE_* vars are only available after doing `mix release` so this would add unnecessary complexity in the documentation
* we also follow livebook behaviour, see this PR   https://github.com/livebook-dev/livebook/pull/1672

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -25,7 +25,9 @@ JF_HOST=localhost:8080
 # JF_NODE_NAME=jellyfish1@127.0.0.1
 
 # Used for grouping Jellyfishes in a cluster
-# JF_COOKIE=jellyfishclustercookie
+# Use different cookies when you want to form 
+# multiple clusters running on shared machines.
+# JF_COOKIE=panuozzo-pollo-e-pancetta
 
 # List of other Jellyfishes we should try to connect to form a cluster
 # JF_NODES="jellyfish2@127.0.0.1 jellyfish3@127.0.0.1"
--- a/.env.sample
+++ b/.env.sample
@@ -40,7 +40,7 @@ JF_HOST=localhost:8080
 # Used for grouping Jellyfishes in a cluster
 # Use different cookies when you want to form 
 # multiple clusters running on shared machines.
-# JF_DIST_COOKIE=panuozzo-pollo-e-pancetta
+# JF_DIST_COOKIE=jellyfish_cookie
 
 # List of other Jellyfishes we should try to connect to form a cluster
 # JF_DIST_NODES="jellyfish2@127.0.0.1 jellyfish3@127.0.0.1"
--- a/.env.sample
+++ b/.env.sample
@@ -20,16 +20,16 @@ JF_HOST=localhost:8080
 
 # Node name used in a cluster.
 # The first part can be any string.
-# The second part has to be Jellyfish address # it is accessible on.
-# JF_NODE_NAME=jellyfish1@127.0.0.1
+# The second part has to be Jellyfish address it is accessible on.
+# JF_DIST_NODE_NAME=jellyfish1@127.0.0.1
 
 # Used for grouping Jellyfishes in a cluster
 # Use different cookies when you want to form 
 # multiple clusters running on shared machines.
-# JF_COOKIE=panuozzo-pollo-e-pancetta
+# JF_DIST_COOKIE=panuozzo-pollo-e-pancetta
 
 # List of other Jellyfishes we should try to connect to form a cluster
-# JF_NODES="jellyfish2@127.0.0.1 jellyfish3@127.0.0.1"
+# JF_DIST_NODES="jellyfish2@127.0.0.1 jellyfish3@127.0.0.1"
 
 # Ports used by Jellyfish when forming a cluster.
 # Besides them, Jellyfish also uses one more service
@@ -63,7 +63,7 @@ JF_WEBRTC_USED=true
 
 # TURN default configuration
 # note: loopback address as INTEGRATED_TURN_IP cannot be used inside a Docker container
-# note: when running locally, JF_INTEGRATED_TURN_IP can be your private ip address 
-JF_INTEGRATED_TURN_IP=<your_public_ip_address>
-JF_INTEGRATED_TURN_LISTEN_IP=0.0.0.0
-JF_INTEGRATED_TURN_PORT_RANGE=50000-50050
+# note: when running locally, JF_WEBRTC_TURN_IP can be your private ip address 
+JF_WEBRTC_TURN_IP=<your_public_ip_address>
+JF_WEBRTC_TURN_LISTEN_IP=0.0.0.0
+JF_WEBRTC_TURN_PORT_RANGE=50000-50050
--- a/.env.sample
+++ b/.env.sample
@@ -15,6 +15,21 @@ JF_HOST=localhost:8080
 # JF_METRICS_IP=0.0.0.0
 # JF_METRICS_PORT=9568
 
+# Defines whether to run Jellyfish in a cluster
+# JF_DIST_ENABLED=false
+
+# Node name used in a cluster.
+# The first part can be any string.
+# The second part has to be Jellyfish address
+# it is accessible on
+# JF_NODE_NAME=jellyfish1@127.0.0.1
+
+# Used for grouping Jellyfishes in a cluster
+# JF_COOKIE=jellyfishclustercookie
+
+# List of other Jellyfishes we should try to connect to form a cluster
+# JF_NODES="jellyfish2@127.0.0.1 jellyfish3@127.0.0.1"
+
 # Token used for authorizing HTTP requests
 JF_SERVER_API_TOKEN=jellyfish_docker_token
 
--- a/.env.sample
+++ b/.env.sample
@@ -48,6 +48,8 @@ JF_HOST=localhost:8080
 # Ports used by Jellyfish when forming a cluster.
 # Besides them, Jellyfish also uses one more service
 # called EPMD, which uses port 4369.
+# Note: those variables only take an effect in docker or
+# after doing `mix release`.
 # JF_DIST_MIN_PORT=9000
 # JF_DIST_MAX_PORT=9000
 
--- a/.env.sample
+++ b/.env.sample
@@ -19,9 +19,23 @@ JF_HOST=localhost:8080
 # JF_DIST_ENABLED=false
 
 # Node name used in a cluster.
-# The first part can be any string.
-# The second part has to be Jellyfish address it is accessible on.
+# It consists of two parts - nodename@hostname.
+# The first part identifies a node on a single machine and can
+# be any string.
+# The second part identifies the host machine and has to be an 
+# ip address or FQDN of a machine Jellyfish runs on.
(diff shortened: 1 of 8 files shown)


.env.sample
```diff
@@ -15,6 +15,21 @@ JF_HOST=localhost:8080
...
 # JF_METRICS_PORT=9568
 
+# Defines whether to run Jellyfish in a cluster
+# JF_DIST_ENABLED=false
+
+# Node name used in a cluster.
+# The first part can be any string.
+# The second part has to be Jellyfish address
+# it is accessible on
+# JF_NODE_NAME=jellyfish1@127.0.0.1
```
Rados13, line 25: I was wondering whether we should describe here that if you want to run all jellyfishes inside one docker network you could provide a simple service name. I think that this is worth stating because to access the container through 127.0.0.1 I think you have to export the EPMD port and multiple instances can't export it at the same time, but you could verify it.

lib/jellyfish/application.ex
```diff
@@ -51,4 +51,50 @@ defmodule Jellyfish.Application do
...
+        {:error, reason} ->
+          raise "Couldn't start Jellyfish node, reason: #{inspect(reason)}"
+      end
+
+      Node.set_cookie(dist_config[:cookie])
+    end
+
+    topologies = [
+      epmd_cluster: [
+        strategy: Cluster.Strategy.Epmd,
```
Rados13, line 75: Shouldn't cluster strategy be set up in the config instead of `application.ex`?

Rados13 (commented):
As we are making that many change in configuration, maybe we should also add possibility to change the port on which epmd runs `ERL_EPMD_PORT` env (more about epmd envs [here](https://www.erlang.org/doc/man/epmd.html#environment_variables)). Other variables worth setting IMO are `inet_dist_listen_min` and `inet_dist_listen_max`. We already use these variables in jellyfish_videoroom, but we configure it by passing the option to elixir, which passes it as an option to erlang. Both of these changes we could add in the next PR.

.env.sample
sgfn: While I am against expanding the comments in .env files even more, I agree this should be mentioned somewhere

lib/jellyfish/config_reader.ex
```diff
@@ -48,19 +50,55 @@ defmodule Jellyfish.ConfigReader do
...
     if value = System.get_env(env) do
       String.downcase(value) not in ["false", "f", "0"]
     end
   end
+
+  def read_dist_config() do
+    if read_boolean("JF_DIST_ENABLED") do
+      node_name_value = System.get_env("JF_NODE_NAME")
+      cookie_value = System.get_env("JF_COOKIE")
+      nodes_value = System.get_env("JF_NODES") || ""
```
sgfn, line 63: Nitpick
```suggestion
      nodes_value = System.get_env("JF_NODES", "")
```

lib/jellyfish/config_reader.ex
```diff
@@ -48,19 +50,55 @@ defmodule Jellyfish.ConfigReader do
...
+      cookie = parse_cookie(cookie_value)
+      nodes = parse_nodes(nodes_value)
+
+      if nodes == [] do
+        Logger.warning("""
+        JF_DIST_ENABLED has been set but JF_NODES remains unset.
+        Make sure that at least one of your Jellyfish instances
+        has JF_NODES set.
+        """)
+      end
```
sgfn, line 88: I'm not saying we should raise here, but is there a case where we'd want to leave JF_NODES empty in a Jellyfish meant to be run in a cluster, while other Jellyfishes in said cluster have JF_NODES set 'correctly'?

.env.sample
```diff
@@ -15,6 +15,21 @@ JF_HOST=localhost:8080
 # JF_METRICS_IP=0.0.0.0
 # JF_METRICS_PORT=9568
 
+# Defines whether to run Jellyfish in a cluster
+# JF_DIST_ENABLED=false
+
+# Node name used in a cluster.
+# The first part can be any string.
+# The second part has to be Jellyfish address
+# it is accessible on
```
sgfn, line 24:
```suggestion
# The second part has to be the address Jellyfish is accessible on
```

rel/env.sh.eex
```diff
@@ -0,0 +1,7 @@
+# release always starts in a distributed mode
+# but to provide unified way of configuring Jellyfish
+# distribution both in the local and production environment
+# we introduce our own env vars and use them to
+# set RELASE_NODE and RELASE_COOKIE
```
sgfn, line 5:
```suggestion
# We introduce our own env vars and use them to set
# RELEASE_NODE and RELEASE_COOKIE.
#
# This is to provide a unified way of configuring Jellyfish distribution
# in both development and production environments
```

sgfn (commented):
A general idea I have is to group all of the distribution env variables under the same prefix, e.g.
- `JF_DIST_ENABLED` -> no change
- `JF_NODE_NAME` -> `JF_DIST_NODE_NAME`
- `JF_COOKIE` -> `JF_DIST_COOKIE`
- `JF_NODES` -> `JF_DIST_NODES`

IMO this isn't overly verbose, and would clearly communicate that these variables are meant to be used together, and that users need to pay attention to them *only* if they wish to run Jellyfish in a cluster.

lib/jellyfish/config_reader.ex
mickel8: That's correct setup. In fact, first Jellyfish very often won't be able to establish connection to other JFs as they are not alive yet

mickel8:
> JF_DIST_ENABLED -> no change
> JF_NODE_NAME -> JF_DIST_NODE_NAME
> JF_COOKIE -> JF_DIST_COOKIE
> JF_NODES -> JF_DIST_NODES


Can do that but what about WebRTC or RTSP? I think we should prefix them too

lib/jellyfish/application.ex
mickel8: I don't think so? What's the benefit?

.env.sample
mickel8: I am not sure we should encourage people to do such setups :thinking:

sgfn:
> @sgfn Can do that but what about WebRTC or RTSP? I think we should prefix them too

Yeah, I've no problem with that, we should go for it

lib/jellyfish/application.ex
Rados13: The main benefit I see is that we will have a clear separation of configuring and using configuration. Currently when someone wants to see how we use `libcluster` have to go to two files `application.ex` to see strategy and to `config_reader.ex` to see how hosts are represented. Also moving strategy to config will make it much easier for future extending to different strategies.

.env.sample
Rados13: I agree we shouldn't encourage people to do so, but this setup is the best way to test distributed jellyfish on a single machine.

mickel8: Ok I missunderstood at first, I thought we want to run everything in one Docker.

I will describe this but I am not sure yet where

lib/jellyfish/application.ex
mickel8: I am not sure

> Currently when someone wants to see how we use libcluster have to go to two files application.ex to see strategy and to config_reader.ex to see how hosts are represented.

If we move to the config.exs we will still have two or even three places: configreader, config.exs, application.ex

>  Also moving strategy to config will make it much easier for future extending to different strategies.

That's a nitpick, I think that keeping everything in an application.ex is easier to read at this stage

roznawsk:
+1 from me for @sgfn comments - other than that LGTM

lib/jellyfish/application.ex
Rados13: But we can move it to config_reader. The idea is to set configuration in one place e.g config_reader and in application.ex only read set values.

rel/env.sh.eex
mickel8: The info that release always starts in a distributed mode is crucial here

.env.sample
```diff
@@ -20,8 +20,7 @@ JF_HOST=localhost:8080
 
 # Node name used in a cluster.
 # The first part can be any string.
-# The second part has to be Jellyfish address
-# it is accessible on
+# The second part has to be Jellyfish address # it is accessible on.
```
roznawsk, line 23:
```suggestion
# The second part has to be Jellyfish address it is accessible on.
```

roznawsk (approved):

lib/jellyfish/application.ex
mickel8: Sorry but I am not convinced :stuck_out_tongue:   I see your point but I think it's better not to mix this config with env vars reading. At least for now

rel/env.sh.eex
```diff
@@ -1,7 +1,37 @@
...
+if [ "$JF_DIST_MAX_PORT" != "" ] ; then
+    case $RELEASE_COMMAND in
+    start*|daemon*)
+        ELIXIR_ERL_OPTIONS="$ELIXIR_ERL_OPTIONS -kernel inet_dist_listen_max $JF_DIST_MAX_PORT"
+        export ELIXIR_ERL_OPTIONS
+        ;;
+    *)
+        ;;
+    esac
+fi
```
Rados13, line 37: I am unsure about `eex` syntax, but don't you set `MIN_PORT` or `MAX_PORT`? Because in the second, if you override the value from the first if?

Rados13 (approved):
Besides one issue LGTM

lib/jellyfish/application.ex
Rados13: As you wish, it's your responsibility as team leader, and you will have to carry the burden of this decision. 🤣

rel/env.sh.eex
mickel8: Nope, this should append to the already existing ELIXIR_ERL_OPTIONS variable

In case of setting both MIN and MAX port we will have `-kernel inet_dist_listen_min 2222 -kernel inet_dist_listen_max 3333`

sgfn (approved):
Approved, with one small note:

Regarding the `.env.sample` file: I'm still firmly against making the *core* Jellyfish documentation spread out in two or more places -- could you just confirm we will duplicate all of this info in the docs, and **not** have the docs say: "for more info, refer to the `.env.sample` file"?

mickel8:
> Approved, with one small note:
>
>Regarding the .env.sample file: I'm still firmly against making the core Jellyfish documentation spread out in two or more >places -- could you just confirm we will duplicate all of this info in the docs, and not have the docs say: "for more info, >refer to the .env.sample file"?

Sure, I will revisit this once I write official docs

PR #97, opened 2023-10-03
[RTC-345] Always require cookie

mickel8:
Setting a default cookie only for dev is possible but would be a little bit tricky so let's always require it


mickel8:
As distribution can only be used in private network (there is no encryption currently), we will keep current config as is and revisit it once we allow for running distribution globally (across different networks)

PR #102, opened 2023-10-05
Extend possible check origins

Rados13:


lib/jellyfish/config_reader.ex
```diff
@@ -50,7 +50,7 @@ defmodule Jellyfish.ConfigReader do
     end
   end
 
-  def read_boolean(env) do
+  defp parse_boolean(env, other_handler \\ nil) do
```
mickel8, line 53: Let's just extend previous function i.e.

```elixir
def read_boolean(env, fallback) do
```

or modify all exisiting functions to have trailing `!` and add `read_boolean` without `!` e.g.

```elixir
def read_check_origin(env) do
  case read_boolean(env) do
    {:ok, value} -> # ...
    {:error, reason} -> # try to parse domains
  end
 end
 ```

That's how livebook does it https://github.com/livebook-dev/livebook/blob/main/lib/livebook/config.ex#L446
 
Now we have parse_boolean which also reads env variable so we mix responsibilities

File: test/jellyfish/config_reader_test.exs [hunk 2]
```diff
@@ -71,6 +71,29 @@ defmodule Jellyfish.ConfigReaderTest do
...
+      System.put_env(env_name, "false")
+      assert ConfigReader.read_check_origin(env_name) == false
+      System.put_env(env_name, "true")
+      assert ConfigReader.read_check_origin(env_name) == true
+      System.put_env(env_name, "jellyfish.ovh jellyfish2.ovh jellyfish3.ovh")
+
+      assert ConfigReader.read_check_origin(env_name) == [
+               "jellyfish.ovh",
+               "jellyfish2.ovh",
+               "jellyfish3.ovh"
```
mickel8, line 87: Should we also allow for ip addresses?
Will we accept those values?

```elixir
check_origin: [
  "//another.com:888",
  "//*.other.com"
]
```

I took this from Phoenix docs https://hexdocs.pm/phoenix/Phoenix.Endpoint.html#socket/3-common-configuration

mickel8 (changes requested):

blazpie (changes requested):

lib/jellyfish/config_reader.ex
```diff
@@ -59,12 +59,31 @@ defmodule Jellyfish.ConfigReader do
...
+          other_handler.(value)
       end
     end
   end
 
+  def read_check_origin(env) do
+    parse_boolean(env, fn value ->
+      hosts = String.split(value, " ")
+
+      if Enum.all?(hosts, &String.contains?(&1, ".")) do
```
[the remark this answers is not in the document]
Rados13: As for these two additional cases, I wouldn't care about custom MFA because it will be only available for a user if he extends jellyfish by himself. For `:conn`, I can add another case that would match a string like this.

test/jellyfish/config_reader_test.exs
Rados13: I added test cases for this example from Phoenix docs. As for IP addresses, after another person's review, I removed any validation.

mickel8 (approved):

PR #104, opened 2023-10-11
Add webhooks for PeerConnected and PeerDisconnected events

Rados13:


lib/jellyfish/event.ex
```diff
@@ -16,8 +20,32 @@ defmodule Jellyfish.Event do
...
+  def broadcast_server_notification(message, room_id) do
+    topic = :server_notification
     Phoenix.PubSub.broadcast(@pubsub, Atom.to_string(topic), {topic, message})
+
+    content = to_proto({topic, message})
+
+    notification = %ServerMessage{content: content} |> ServerMessage.encode()
+
+    Room.receive_room_notification(room_id, notification)
   end
```
mickel8, line 49: Maybe it would be better to create a separate process that will be responsible for subscribing for events and sending notificaitons using webhooks. This way we wouldn't have to modify event.ex. If neeed we could put webhook url inside Registry

lib/jellyfish/webhook_notifier.ex
```diff
@@ -0,0 +1,68 @@
+defmodule Jellyfish.WebhookNotifier do
+  @moduledoc """
+  Module responsible for sending notifications to webhooks.
```
mickel8, line 3: Sometimes we write `sending notifcations to webhooks` and sometimes  `sending notification through webhooks`. Would be nice to unify this

mickel8 (changes requested):
Nicely done! :fire: I've just added a couple of comments mostly related to the code consistency

lib/jellyfish/room.ex
```diff
@@ -376,6 +382,9 @@ defmodule Jellyfish.Room do
     :ok
   end
 
+  def registry_id(room_id),
+    do: {:via, Registry, {Jellyfish.RoomRegistry, room_id}}
```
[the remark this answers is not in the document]
Rados13: During testing another approach with storing webhook_url in Registry, I found out that we create [`registry_id` from hand in `RoomService`](https://github.com/jellyfish-dev/jellyfish/blob/a4dd56e4efa4a241192f178c90c3bc67e95d0f2b/lib/jellyfish/room_service.ex#L225), which is error-prone because if we change how registry_id is made in `Room` we could forget to change it in `RoomService`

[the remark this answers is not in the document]
mickel8: Ah, now I see, it became public. Let's move this then to the top, before `start` function as now it's part of the public API

test/jellyfish_web/integration/server_socket_test.exs
```diff
@@ -22,10 +22,14 @@ defmodule JellyfishWeb.Integration.ServerSocketTest do
   }
 
   alias JellyfishWeb.{PeerSocket, ServerSocket, WS}
+  alias Phoenix.PubSub
 
   @port 5907
+  @webhook_port 2137
```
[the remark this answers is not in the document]
mickel8: Yeah yeah, I know this meme but maybe we should not mixin political memes in the code? I am not sure :thinking: Maybe use panuozzo pollo e pancetta price x2 e.g. 2929

lib/jellyfish_web/controllers/room_controller.ex
```diff
@@ -82,10 +82,14 @@ defmodule JellyfishWeb.RoomController do
       |> render("show.json", room: room, jellyfish_address: jellyfish_address)
     else
       {:error, :invalid_max_peers} ->
-        {:error, :bad_request, "maxPeers must be a number"}
+        max_peers = Map.get(params, "maxPeers")
+        {:error, :bad_request, "maxPeers must be a number received maxPeers was: #{max_peers}"}
```
mickel8, line 86: Let's write all of those error messages as:

```suggestion
        {:error, :bad_request, "Expected maxPeers to be a number, got: : #{max_peers}"}
```

lib/jellyfish_web/peer_socket.ex
```diff
@@ -130,14 +130,10 @@ defmodule JellyfishWeb.PeerSocket do
   @impl true
   def terminate(_reason, state) do
```
mickel8, line 131: I would get rid of this function at all as it won't be called in every case so sometimes we get this log and soimetimes not

test/support/ws.ex
```diff
@@ -11,6 +11,11 @@ defmodule JellyfishWeb.WS do
     WebSockex.start_link(url, __MODULE__, state)
   end
 
+  def start(url, type) do
+    state = %{caller: self(), type: type}
+    WebSockex.start(url, __MODULE__, state)
+  end
```
mickel8, line 17: Nitpick: put this before start_link

test/jellyfish_web/integration/server_socket_test.exs
```diff
@@ -167,33 +184,82 @@ defmodule JellyfishWeb.Integration.ServerSocketTest do
...
+    assert_receive %RoomDeleted{room_id: ^room_id}
+
+    assert_receive {:room_deleted, %RoomDeleted{room_id: ^room_id}},
+                   1_000
+
+    refute_received %PeerDisconnected{room_id: ^room_id, peer_id: ^peer_id}
+    refute_received {:peer_disconnected, %PeerDisconnected{room_id: ^room_id, peer_id: ^peer_id}}
+  end
+
+  test "sends a message when peer connects and room crashes", %{conn: conn} do
```
mickel8, line 212: merge this with `sends a message when room crashes`

test/jellyfish_web/integration/server_socket_test.exs
```diff
@@ -167,33 +184,82 @@ defmodule JellyfishWeb.Integration.ServerSocketTest do
...
+
+    assert_receive %PeerDisconnected{room_id: ^room_id, peer_id: ^peer_id}
+
+    assert_receive {:peer_disconnected, %PeerDisconnected{room_id: ^room_id, peer_id: ^peer_id}},
+                   2_000
+
+    delete(conn, ~p"/room/#{room_id}")
+  end
+
+  def prepare_for_server_notification_test(conn) do
```
mickel8, line 246: This function name is too generic :thinking:

test/support/webhook_plug.ex
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def call(conn, _opts) do
+    {:ok, body, conn} = Plug.Conn.read_body(conn, [])
+    notification = Jason.decode!(body)
+
+    notification =
+      notification |> Map.get("notification") |> ServerMessage.decode() |> Map.get(:content)
+
+    :ok = PubSub.broadcast(@pubsub, "webhook", notification)
```
mickel8, line 22: Instead of broadcasting `{:peer_crashed, %PeerCrashed{}}`, let's broadcast `{:webhook_notification, %PeerCrashed{}}`.

test/jellyfish_web/integration/server_socket_test.exs
```diff
@@ -22,10 +22,14 @@ defmodule JellyfishWeb.Integration.ServerSocketTest do
   }
```
mickel8, line 23: We should rename this file and module to e.g. `ServerNotificationTest`?

test/jellyfish_web/integration/server_socket_test.exs
```diff
@@ -167,33 +184,82 @@ defmodule JellyfishWeb.Integration.ServerSocketTest do
...
+    assert_receive {:peer_connected, %PeerConnected{peer_id: ^peer_id, room_id: ^room_id}}, 1_000
 
-    conn = delete(conn, ~p"/room/#{room_id}/")
-    assert response(conn, :no_content)
-
-    assert_receive %PeerDisconnected{peer_id: ^peer_id, room_id: ^room_id}
+    {room_id, peer_id, conn}
   end
 
   def create_and_authenticate() do
```
mickel8, line 265: Could we move "test metrics" next to other tests?

mickel8 (changes requested):

lib/jellyfish_web/peer_socket.ex
Rados13: This function is required by behavior `Phoenix.Socket.Transport`, but if you want I can remove this log from it.

lib/jellyfish/room_service.ex
```diff
@@ -17,7 +16,7 @@ defmodule Jellyfish.RoomService do
   @spec find_room(Room.id()) :: {:ok, pid()} | {:error, :room_not_found}
   def find_room(room_id) do
     case Registry.lookup(Jellyfish.RoomRegistry, room_id) do
-      [{room_pid, nil}] ->
+      [{room_pid, _webhook_url}] ->
```
Karolk99, line 19: I would change the `_webhook_url` to a different name as we never assign webhook to registry id.

Karolk99 (approved):

lib/jellyfish_web/peer_socket.ex
mickel8: Yeah, let's do this

PR #106, opened 2023-10-17
Add HTTP requests metrics

sgfn:
Metrics written for the HLS stress tests.

I'm open to discussion whether we want to keep all, some or none of them

### DIFF
--- a/lib/jellyfish_web/traffic_metrics_plug.ex
+++ b/lib/jellyfish_web/traffic_metrics_plug.ex
@@ -1,0 +1,55 @@
+defmodule JellyfishWeb.TrafficMetricsPlug do
+  @moduledoc false
+
+  import Plug.Conn
+
+  # Assuming additional 14 bytes per request, e.g.
+  #    V     VVVVVVVVVV V            V V
+  # GET /room HTTP/1.1\r\n   [...]   \r\n
+  @request_add 14
+
+  # Assuming additional 19 bytes per response, e.g.
+  # VVVVVVVVVVVVVVVV V            V V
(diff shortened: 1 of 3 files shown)


lib/jellyfish_web/traffic_metrics_plug.ex
```diff
@@ -0,0 +1,55 @@
+defmodule JellyfishWeb.TrafficMetricsPlug do
+  @moduledoc false
+
+  import Plug.Conn
+
+  # Assuming additional 14 bytes per request, e.g.
```
Karolk99, line 6: Everything seems legit. Could you tell me why you chose to write TrafficMetricsPlug like that? I mean those constants

Rados13 (approved):

lib/jellyfish_web/traffic_metrics_plug.ex
sgfn: I couldn't find an easier way to get the exact (or, in this case, approximate) response and request sizes. These constants are estimates which, as far as I've tested, work reasonably well.

The logic is as follows:
- We can calculate part of the request/response size by adding `byte_size()` of e.g. the body, the method, each `{key, value}` pair from the headers
- But that doesn't count everything -- for instance, header syntax is `Key: Value\r\n`, and we only count `byte_size("Key") + byte_size("Value")` (in the comments, `V` directly above a character means it was not counted)
- That's where the constants come from, from the example above you can see we're missing 4 bytes per header, and so on.

I assumed the most common case which most HTTP clients should adhere to

mickel8 (approved):

Karolk99 (approved):

PR #114, opened 2023-11-02
[RTC-379] Fix "list all rooms" test. Make RoomControllerTest sync.

Rados13:

This PR fixes two tests:
- Metrics notification tests by increasing timeouts because sometimes there was a situation that the notification wasn't received
- `list all rooms` test by disallowing parallel tests for this module and deleting all rooms before running any test case in this module. Previously, there was a possibility that there was another room in jellyfish before starting this test case.

In addition, in `component_controller_test.exs`, the naming for connection was unified, so in all test cases, we use the name `conn` instead of mixing `conn` and `room_conn`

### DIFF
--- a/test/jellyfish_web/controllers/room_controller_test.exs
+++ b/test/jellyfish_web/controllers/room_controller_test.exs
@@ -1,5 +1,5 @@
 defmodule JellyfishWeb.RoomControllerTest do
-  use JellyfishWeb.ConnCase
+  use JellyfishWeb.ConnCase, async: false
 
   import OpenApiSpex.TestAssertions
   alias Jellyfish.RoomService
@@ -9,6 +9,8 @@ defmodule JellyfishWeb.RoomControllerTest do
   setup %{conn: conn} do
     server_api_token = Application.fetch_env!(:jellyfish, :server_api_token)
     conn = put_req_header(conn, "authorization", "Bearer " <> server_api_token)
+
+    delete_all_rooms(conn)
 
     on_exit(fn -> delete_all_rooms(conn) end)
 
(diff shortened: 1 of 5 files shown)


sgfn (approved):

mickel8:
I am not sure but we could make `room_conn` and `conn` distinction on purpose. @Karolk99 do you remember why?

test/jellyfish_web/controllers/room_controller_test.exs
```diff
@@ -10,6 +10,8 @@ defmodule JellyfishWeb.RoomControllerTest do
     server_api_token = Application.fetch_env!(:jellyfish, :server_api_token)
     conn = put_req_header(conn, "authorization", "Bearer " <> server_api_token)
 
+    delete_all_rooms(conn)
```
mickel8, line 13: What about going even further and raise if some other test didn't do cleanup correctly? Does it make any sense to you? If not I would just move `delete_all_rooms(conn)` into `setup_all` so that we call it only once, before executing any test

```suggestion
setup_all %{conn: conn} do
  conn = get(conn, ~p"/room")
  response = json_response(conn, :ok)
  assert length(response["data"]) == 0
end
```

mickel8 (approved):

Karolk99:
> I am not sure but we could make `room_conn` and `conn` distinction on purpose. @Karolk99 do you remember why?

I'm not sure. I think I made those tests analogical to the setup implementation. Sometimes there is a problem with reusing the connection (Plug.Conn.AlreadySentError). @Rados13 why didn't you change `room_conn` in setup?

test/jellyfish_web/controllers/room_controller_test.exs
Rados13: It makes sense that all tests should clean their state after tests. But it won't be very pleasant to debug this problem in the future, and also, it is very easy to introduce it. Without this fix, 1 on 30 tests were failing, so we easily could overlook that our change introduces RC in tests. It is a little bit problem with this test `list all rooms` that this test could be affected by other tests.

Rados13:
> > I am not sure but we could make `room_conn` and `conn` distinction on purpose. @Karolk99 do you remember why?
> 
> I'm not sure. I think I made those tests analogical to the setup implementation. Sometimes there is a problem with reusing the connection (Plug.Conn.AlreadySentError). @Rados13 why didn't you change `room_conn` in setup?

I didn't change it because of the error you mentioned. If I understand this Phoenix mechanism correctly, we should use the output connection from the previous request as input in the next request. But because `on_exit` is called after tests, we can't use it that way. But I am not sure about it.

Karolk99 (approved):

PR #116, opened 2023-11-08
[RTC-380] Revisit Jellyfish logs

Rados13:

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -19,6 +19,20 @@ config :logger, :console,
   metadata: [:request_id, :room_id]
 
 config :phoenix, :json_library, Jason
+config :phoenix, :logger, false
+
+config :logger,
+  compile_time_purge_matching: [
+    [module: Membrane.RTC.Engine, level_lower_than: :warning],
+    [module: Membrane.RTC.Engine.Endpoint.RTSP, level_lower_than: :warning],
+    [module: Membrane.RTC.Engine.Endpoint.HLS, level_lower_than: :warning],
+    [module: Membrane.RTC.Engine.Endpoint.WebRTC, level_lower_than: :warning],
+    [module: Membrane.RTC.Engine.Tee, level_lower_than: :warning],
+    [
+      module: Membrane.RTC.Engine.Endpoint.WebRTC.RTPConnectionAllocator,
+      level_lower_than: :warning
+    ]
+  ]
 
 config :jellyfish,
   divo: "docker-compose.yaml",
--- a/lib/jellyfish/room.ex
+++ b/lib/jellyfish/room.ex
@@ -15,6 +15,7 @@ defmodule Jellyfish.Room do
   alias Membrane.ICE.TURNManager
   alias Membrane.RTC.Engine
   alias Membrane.RTC.Engine.Message
+  alias Membrane.RTC.Engine.Message.{EndpointAdded, EndpointRemoved}
 
   @enforce_keys [
     :id,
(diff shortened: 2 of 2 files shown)


config/config.exs
```diff
@@ -19,6 +19,20 @@ config :logger, :console,
...
+  compile_time_purge_matching: [
+    [module: Membrane.RTC.Engine, level_lower_than: :warning],
+    [module: Membrane.RTC.Engine.Endpoint.RTSP, level_lower_than: :warning],
+    [module: Membrane.RTC.Engine.Endpoint.HLS, level_lower_than: :warning],
+    [module: Membrane.RTC.Engine.Endpoint.WebRTC, level_lower_than: :warning],
+    [module: Membrane.RTC.Engine.Tee, level_lower_than: :warning],
+    [
+      module: Membrane.RTC.Engine.Endpoint.WebRTC.RTPConnectionAllocator,
+      level_lower_than: :warning
+    ]
```
mickel8, line 34: * Will we log now information about new incoming tracks? If engine emits such message we could catch it in Jellyfish and log it on our side instead of relying on Engine
* AFAIK there should be an option to remove all logs from specific application so that we don't have to list all modules
* I think that some of the logs in the Engine like those related to simulcast bandwidth estimation should be moved to debug anyway

lib/jellyfish/room.ex
```diff
@@ -365,6 +366,31 @@ defmodule Jellyfish.Room do
     {:noreply, state}
   end
 
+  @impl true
+  def handle_info(%Membrane.RTC.Engine.Message.EndpointMessage{} = msg, state) do
```
mickel8, line 370: alias this too

lib/jellyfish/room.ex
```diff
@@ -365,6 +366,31 @@ defmodule Jellyfish.Room do
...
+             not is_map_key(state.peers, endpoint_id) do
+    {:noreply, state}
+  end
+
+  @impl true
+  def handle_info(
+        %EndpointAdded{endpoint_id: endpoint_id},
+        state
+      )
+      when is_map_key(state.components, endpoint_id) or is_map_key(state.peers, endpoint_id) do
```
mickel8, line 390: Isn't this RC? When we add and remove an endpoint fast enough we can end up removing endpoint faster than reading a message from the engine? I am not sure but I would remove those guards and create one handle_info for all notifications (if possible)

Rados13: We always add/remove peers/components before receiving a notification about it, so there is no RC. This is also why we had logs `Unexpected info RoomAdded/RoomDeleted`. It would be hard to have one `handle_info` for all notifications as for example we handle `Message.EndpointMessage` with `media_events` separately.

config/config.exs
Rados13: As for the last issue [PR for this is created](https://github.com/jellyfish-dev/membrane_rtc_engine/pull/327).
As for other issues they are applied.

lib/jellyfish/room.ex
```diff
@@ -365,6 +366,37 @@ defmodule Jellyfish.Room do
...
+    {:noreply, state}
+  end
+
+  @impl true
+  def handle_info(%Message.TrackAdded{} = track_info, state) do
```
Karolk99, line 395: What about `TrackRemoved` message?

config/config.exs
```diff
@@ -19,6 +19,13 @@ config :logger, :console,
   metadata: [:request_id, :room_id]
 
 config :phoenix, :json_library, Jason
+config :phoenix, :logger, false
+
+config :logger,
+  compile_time_purge_matching: [
+    [application: :membrane_rtc_engine, level_lower_than: :warning],
+    [application: :membrane_rtc_engine_webrtc, level_lower_than: :warning]
```
Karolk99, line 27: Why only webrtc is included? What about hls and rtsp?

lib/jellyfish/room.ex
```diff
@@ -365,6 +366,37 @@ defmodule Jellyfish.Room do
...
+    {:noreply, state}
+  end
+
+  @impl true
+  def handle_info(
+        %EndpointRemoved{endpoint_id: endpoint_id},
+        state
+      )
+      when not is_map_key(state.components, endpoint_id) or
+             not is_map_key(state.peers, endpoint_id) do
```
sgfn, line 381: I'm not a big fan of this solution, but if we're keeping it:

1. per De Morgan, I think there should be an `and` there
2. how about putting it in a `defguardp` or something? we have the same condition twice, one time it's negated

sgfn (commented):
I concur with the other review comments

lib/jellyfish/room.ex
```diff
@@ -365,6 +366,37 @@ defmodule Jellyfish.Room do
     {:noreply, state}
   end
 
+  @impl true
+  def handle_info(%Message.EndpointMessage{} = msg, state) do
```
mickel8, line 370:
```suggestion
  def handle_info(%EndpointMessage{} = msg, state) do
```

mickel8 (approved):
Just fix other comments and IMO we can merge

config/config.exs
Rados13: I first thought there were not many logs from HLS and RTSP, so it isn't necessary to remove them, but maybe that is the right direction to be consistent.

sgfn (approved):

Karolk99 (approved):

PR #131, opened 2023-12-14
HLS: subscribe for peers and components instead of specific tracks

Karolk99:

### DIFF
--- a/lib/jellyfish_web/api_spec/subscription.ex
+++ b/lib/jellyfish_web/api_spec/subscription.ex
@@ -3,20 +3,7 @@ defmodule JellyfishWeb.ApiSpec.Subscription do
 
   alias OpenApiSpex.Schema
 
-  defmodule Origin do
-    @moduledoc false
-
-    require OpenApiSpex
-
-    OpenApiSpex.schema(%{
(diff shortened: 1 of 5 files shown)


lib/jellyfish_web/api_spec/subscription.ex
```diff
@@ -26,10 +26,11 @@ defmodule JellyfishWeb.ApiSpec.Subscription do
...
       type: :object,
       properties: %{
-        tracks: %Schema{
+        origins: %Schema{
           type: :array,
-          description: "List of tracks that hls endpoint will subscribe for",
-          items: Track
+          description:
+            "List of peers and components whose tracks the HLS endpoint will subscribe to",
+          items: Origin
```
mickel8, line 33: I would just convert this to string. Also module name needs an update

openapi.yaml
```diff
@@ -5,25 +5,126 @@
 components:
```
mickel8, line 5: Pretty weird that there are so many changes but I checked and this is indeed how openapi.yaml is generated on this branch

mickel8 (approved):

PR #135, opened 2023-12-27
Refactor file and rtsp component tests

Karolk99:

Some tests were removed instead of moving them to controller tests because they were redundant also some tests stayed in `jellyfish/component` folder because they don't return any properties.


roznawsk (commented):
LGTM, although what do you think about removing file_test and rtsp_test altogether? They seem kindof useless now

roznawsk (approved):

PR #136, opened 2023-12-27
Rtc 433 update openapi

roznawsk:
 - [x ] I hereby confirm that a Pull Request involving updates to the Software Development Kit (SDK) has been smoothly merged, currently awaits processing, or is otherwise deemed unnecessary in this context.
 - [x ] I also affirm that another Pull Request, specifically addressing updates to the documentation body (commonly referred to as 'docs'), has either been successfully incorporated, is in the process of review, or is considered superfluous under the prevailing circumstances.

### DIFF
--- a/lib/jellyfish_web/api_spec/component/file.ex
+++ b/lib/jellyfish_web/api_spec/component/file.ex
@@ -19,7 +19,8 @@ defmodule JellyfishWeb.ApiSpec.Component.File do
           description: "Path to track file. Must be either OPUS encapsulated in Ogg or raw h264",
           example: "/root/video.h264"
         }
-      }
+      },
+      required: [:filePath]
     })
   end
 
--- a/lib/jellyfish_web/api_spec/component/file.ex
+++ b/lib/jellyfish_web/api_spec/component/file.ex
@@ -16,8 +16,8 @@ defmodule JellyfishWeb.ApiSpec.Component.File do
       properties: %{
         filePath: %Schema{
           type: :string,
-          description: "Path to track file. Must be either OPUS encapsulated in Ogg or raw h264",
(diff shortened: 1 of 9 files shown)


sgfn (approved):

lib/jellyfish_web/api_spec/component/file.ex
```diff
@@ -13,7 +13,14 @@ defmodule JellyfishWeb.ApiSpec.Component.File do
       title: "ComponentPropertiesFile",
       description: "Properties specific to the File component",
       type: :object,
-      properties: %{}
+      properties: %{
+        filePath: %Schema{
+          type: :string,
+          description: "Path to track file. Must be either OPUS encapsulated in Ogg or raw h264",
+          example: "/root/video.h264"
```
Karolk99, line 20: Isn't `filePath` a related path? Because `/root/video.h264`, which appears to be an absolute path, could easily be misunderstood."

lib/jellyfish_web/api_spec/component/rtsp.ex
```diff
@@ -13,7 +13,32 @@ defmodule JellyfishWeb.ApiSpec.Component.RTSP do
       title: "ComponentPropertiesRTSP",
       description: "Properties specific to the RTSP component",
       type: :object,
-      properties: %{}
+      properties: %{
+        sourceUri: %Schema{
+          type: :string,
+          description: "URI of RTSP source stream",
+          example: "rtsp://localhost:554/stream"
```
Karolk99, line 20: I wouldn't add an example if this schema is only used to describe a response message. What do you think?

Karolk99 (approved):

PR #143, opened 2024-01-19
[RTC-454] Healthcheck endpoint

sgfn:
About the topic:
https://emmer.dev/blog/writing-meaningful-health-check-endpoints/
https://tyk.io/blog/api-health-checks-how-to-keep-apps-healthy/

There are several things we need to consider:

1. Whether we want to perform some actual checks before responding (+ include some additional data in the response)
    1. We're effectively returning a hard-coded 200 every time, which seems more like a ping endpoint than a healthcheck endpoint...
    2. ...but:
        1. We don't have any microservices or databases we depend on
        2. There is little need for the distinction between *live* and *ready* states of Jellyfish:
            1. When it starts, it should function correctly from the get-go
            2. During its lifetime, we could designate it as *not-ready* whenever the load is too high to handle more traffic, but detecting that from within the app wouldn't be trivial
        3. We already have metrics reporting resource usage, amount of rooms etc. We could consider adding uptime, I guess (?)
2. Whether issuing healthcheck requests should require authentication
    1. IMO right now it isn't necessary, but if we choose to include anything other than `{"status": "UP"}` in the response, we should use auth

Discussion on these topics is welcome 🙂 

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -126,7 +126,7 @@ RUN chmod +x docker-entrypoint.sh
 
 ENV HOME=/app
 
-HEALTHCHECK CMD curl --fail -H "authorization: Bearer ${JF_SERVER_API_TOKEN}" http://localhost:${JF_PORT:-8080}/room || exit 1
+HEALTHCHECK CMD curl --fail -H "authorization: Bearer ${JF_SERVER_API_TOKEN}" http://localhost:${JF_PORT:-8080}/health || exit 1
 
 ENTRYPOINT ["./docker-entrypoint.sh"]
 
--- a/lib/jellyfish_web/api_spec/health_report.ex
+++ b/lib/jellyfish_web/api_spec/health_report.ex
@@ -1,0 +1,29 @@
+defmodule JellyfishWeb.ApiSpec.HealthReport do
+  @moduledoc false
+
+  require OpenApiSpex
+
+  defmodule Status do
+    @moduledoc false
+
+    require OpenApiSpex
+
+    OpenApiSpex.schema(%{
+      title: "HealthReportStatus",
+      description: "Informs about the status of Jellyfish",
+      type: :string,
+      enum: ["UP", "DOWN"],
+      example: "UP"
+    })
+  end
+
+  OpenApiSpex.schema(%{
+    title: "HealthReport",
+    description: "Describes overall Jellyfish health",
+    type: :object,
+    properties: %{
+      status: Status
+    },
+    required: [:status]
+  })
+end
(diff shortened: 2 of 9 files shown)


Rados13 (approved):
Nice research and great attached resources. As for the questions, I think we could add in the healthcheck response information about distribution. If, for some reason, the local node is not alive, we could state that the service is DOWN. The additional information that we could provide in Healthcheck is how many nodes this jellyfish is connected. Adding information about uptime could also be beneficial. I agree with other suggestions.

lib/jellyfish_web/api_spec/health_report.ex
```diff
@@ -0,0 +1,29 @@
...
+  defmodule Status do
+    @moduledoc false
+
+    require OpenApiSpex
+
+    OpenApiSpex.schema(%{
+      title: "HealthReportStatus",
+      description: "Informs about the status of Jellyfish",
+      type: :string,
+      enum: ["UP", "DOWN"],
```
roznawsk, line 15: Fun fact: when adding string enum, we have to ensuer that the values aren't any of:
`'0', 'off', 'f', 'false', 'n', 'no', '1', 'on', 't', 'true', 'y', 'yes'` , since they will be converted to boolean when generating the python SDK. https://docs.pydantic.dev/2.0/usage/types/booleans/

In this case we are perfectly fine.

roznawsk (approved):
Consider adding test for asserting the response from the `/health` endpoint

roznawsk:
I agree with Rados13 on the great research 👍🏽 . We can add information about distribution to the response.

lib/jellyfish_web/api_spec/health_report.ex
sgfn: I had no idea -- perhaps we should leave it as a memo somewhere?

roznawsk: We may, however I am not sure where

Dockerfile
```diff
@@ -126,7 +126,7 @@ RUN chmod +x docker-entrypoint.sh
 
 ENV HOME=/app
 
-HEALTHCHECK CMD curl --fail -H "authorization: Bearer ${JF_SERVER_API_TOKEN}" http://localhost:${JF_PORT:-8080}/room || exit 1
+HEALTHCHECK CMD curl --fail -H "authorization: Bearer ${JF_SERVER_API_TOKEN}" http://localhost:${JF_PORT:-8080}/health || exit 1
```
Rados13, line 129: Do we currently need this authorization?

roznawsk (approved):

Dockerfile
sgfn: Yes, currently `/health` is piped through `:api` in the router

PR #148, opened 2024-02-05
RTC-459 Fix metrics

Rados13:


lib/jellyfish/room_service.ex
```diff
@@ -216,10 +217,22 @@ defmodule Jellyfish.RoomService do
 
     Phoenix.PubSub.broadcast(Jellyfish.PubSub, room_id, :room_crashed)
     Event.broadcast_server_notification({:room_crashed, room_id})
+    clear_room_metrics(room_id)
 
     {:noreply, state}
   end
 
+  defp clear_room_metrics(room_id) do
```
roznawsk, line 225: 👌🏽

test/jellyfish_web/integration/peer_socket_test.exs
```diff
@@ -30,6 +30,9 @@ defmodule JellyfishWeb.Integration.PeerSocketTest do
   end
 
   setup_all do
+    RoomService.list_rooms()
```
roznawsk, line 33: On general note: since we are on a quest to always remove rooms after test, I found that there are some rooms left after running hls component test.
You can run `elixir --no-halt -S mix test test/jellyfish_web/controllers/component/hls_component_test.exs` and then check from dashboard.

I guess if we assure that no test leaves empty room we don't have to remove the rooms here.
Also, AFAIK the tests are async by default, so could this mess with some other tests?

lib/jellyfish/room_service.ex
```diff
@@ -9,8 +9,8 @@ defmodule Jellyfish.RoomService do
 
   alias Jellyfish.{Event, Room, WebhookNotifier}
 
-  # in seconds
-  @metric_interval 10
+  # in milliseconds
+  @metric_interval 10 * Application.compile_env!(:jellyfish, :metrics_scrape_interval)
```
roznawsk, line 13: The scrape interval is multiple of 10 milliseconds, so the default interval of `1000` corresponds to scraping each 10 seconds. I think it's a bit confusing. Maybe the `metrics_scrape_interval` env variable should represent milliseconds? We could add check preventing setting values lower than 10 ms.

lib/jellyfish/room_service.ex
```diff
@@ -216,10 +217,22 @@ defmodule Jellyfish.RoomService do
...
 
     {:noreply, state}
   end
 
+  defp clear_room_metrics(room_id) do
+    :telemetry.execute(
+      [:jellyfish, :room],
+      %{
+        peers: 0,
+        peer_time_total: 0
```
sgfn, line 230: Setting `peer_time_total` here is a no-op, as it is an accumulator

test/jellyfish_web/integration/peer_socket_test.exs
```diff
@@ -171,6 +174,38 @@ defmodule JellyfishWeb.Integration.PeerSocketTest do
...
+  test "proper calculated peer metrics", %{room_id: room_id, token: token, conn: conn} do
+    assert %{} = get_peers_room_metrics()
+    create_and_authenticate(token)
+
+    peers_in_room_key = "jellyfish_room_peers{room_id=\"#{room_id}\"}"
+    peers_in_room_time_key = "jellyfish_room_peer_time_total_seconds{room_id=\"#{room_id}\"}"
+
+    metrics_after_one_tick = %{
+      peers_in_room_key => "1",
+      peers_in_room_time_key => "500",
```
sgfn, line 186: Peer time total is in seconds

lib/jellyfish/room_service.ex
sgfn: This interval is for WebRTC metrics, not Jellyfish metrics. It really shouldn't be used here. Please consult @mickel8 and @Karolk99 on whether we can and should make the interval for Jellyfish metrics configurable.

sgfn (changes requested):

config/config.exs
```diff
@@ -9,7 +9,9 @@ config :jellyfish, JellyfishWeb.Endpoint,
   pubsub_server: Jellyfish.PubSub,
   live_view: [signing_salt: "/Lo03qJT"]
 
-config :jellyfish, metrics_scrape_interval: 1000
+config :jellyfish,
+  webrtc_metrics_scrape_interval: 1000,
+  peer_metrics_scrape_interval: 10
```
sgfn, line 14: Why not jellyfish_metrics_ or room_metrics_?

lib/jellyfish/room.ex
```diff
@@ -458,6 +465,12 @@ defmodule Jellyfish.Room do
     {:noreply, state}
   end
 
+  @impl true
+  def handle_info(%EndpointRemoved{} = track_info, state) do
+    Logger.info("Endpoint removed #{track_info.endpoint_id} removed track #{inspect(track_info)}")
```
sgfn, line 470: I don't see any track info in the EndpointRemoved struct -- was this what you meant to do?

```suggestion
  @impl true
  def handle_info(%TrackRemoved{} = track_info, state) do
    Logger.info("Endpoint #{track_info.endpoint_id} removed track #{inspect(track_info)}")
```

lib/jellyfish/room.ex
```diff
@@ -437,6 +437,13 @@ defmodule Jellyfish.Room do
     {:noreply, state}
   end
 
+  @impl true
+  def handle_info(%EndpointRemoved{endpoint_id: endpoint_id}, state) do
+    {_endpoint, state} = pop_in(state, [:peers, endpoint_id])
+    Logger.info("Peer #{endpoint_id} removed")
+    {:noreply, state}
+  end
```
sgfn, line 445: Also, we already have this case above? Or am I missing something?

Rados13: Yup, as I had a few breaks between writing both cases, I forgot about the previous one.

lib/jellyfish/room.ex
Rados13: Almost I was simply lazy. I duplicated lines above and forgot to change them 😅 .

sgfn (approved):
👍

roznawsk (approved):

PR #150, opened 2024-02-07
[RTC-456] Add peerless room purge

sgfn:


test/jellyfish_web/controllers/room_controller_test.exs
```diff
@@ -118,22 +118,6 @@ defmodule JellyfishWeb.RoomControllerTest do
...
-      response = json_response(conn, :ok)
-      assert_response_schema(response, "RoomDetailsResponse", @schema)
-
-      assert %{
-               "id" => ^id,
-               "config" => %{},
-               "components" => [],
-               "peers" => []
-             } = response["data"]
-    end
```
Rados13, line 135: Why did you remove this test?

Rados13 (approved):

test/jellyfish_web/controllers/room_controller_test.exs
sgfn: It doesn't add anything of value, the checks performed here are already done in "renders room when data is valid" and "renders room when data is valid, custom room_id"

test/jellyfish/room_test.exs
```diff
@@ -0,0 +1,55 @@
+defmodule Jellyfish.RoomTest do
+  use ExUnit.Case, async: true
+
+  alias Jellyfish.{Peer, Room}
+
+  @purge_timeout_s 1
```
mickel8, line 6: The way those tests are written right now will cause us to wait a couple of seconds until they are executed. I am wondering if we could make this better. We can discuss this offline

lib/jellyfish/room.ex
```diff
@@ -554,6 +560,19 @@ defmodule Jellyfish.Room do
...
   end
 
+  @impl true
+  def handle_info(:peerless_purge, state) do
+    if peerless_long_enough?(state) do
+      Logger.info(
+        "Removing room because it was peerless for #{state.config.peerless_purge_timeout} seconds"
+      )
+
+      {:stop, :normal, state}
```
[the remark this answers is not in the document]
sgfn: But that's exactly how `RoomService` removes rooms -- [here](https://github.com/jellyfish-dev/jellyfish/blob/a403ecc47e50da2163fcdf08f8e333ff607ec4d4/lib/jellyfish/room_service.ex#L263)

We're manually removing all of the linked processes in the `terminate` callback, no?

test/jellyfish_web/controllers/room_controller_test.exs
Rados13: Practically speaking, yes; philosophically, no. It is not the responsibility of these tests to verify that action X occurs, especially if it's not mentioned in their names. However, it is easy to accidentally refactor these tests and unintentionally remove coverage for this edge case. If you insist on removing this test, I suggest adding a note to one of the other tests indicating that it also covers the situation previously handled by the removed test.

Rados13: Practically speaking, yes; philosophically, no. It is not the responsibility of these tests to verify that action X occurs, especially if it's not mentioned in their names. However, it is easy to accidentally refactor these tests and unintentionally remove coverage for this edge case. If you insist on removing this test, I suggest adding a note to one of the other tests indicating that it also covers the situation previously handled by the removed test or modifying their names.

Rados13 (approved):

test/jellyfish_web/controllers/room_controller_test.exs
sgfn: Updated name of test

sgfn:
Bumped Elixir to 1.15 because we use Map.intersect which was added in that version

mickel8 (approved):

PR #152, opened 2024-02-09
Properly remove component from state when crashed

gfodor:
This PR updates the room genserver to properly remove a component entry when its endpoint crashes, and also incorporates the recent changes to introduce helper functions for this. The one thing I wasn't sure about is the fact that the track removeal events will go out in the case of a peer crash.


roznawsk (approved):
LGTM!

We expect JF to send `TrackRemoved` notifications even if the Peer crashed. In general we expect each notification indicating allocating a resource (e.g. `PeerConnected`, `TrackAdded`) to have a corresponding notification after the resource is removed (`PeerDisconnected`, `TrackRemoved`)

mickel8:
Awesome, thanks!

PR #170, opened 2024-03-28
Add JF_S3_PATH_PREFIX to config and restrict s3_config to be provided…

Karolk99:
… through only one method

### DIFF
--- a/lib/jellyfish_web/controllers/component_controller.ex
+++ b/lib/jellyfish_web/controllers/component_controller.ex
@@ -83,6 +83,14 @@ defmodule JellyfishWeb.ComponentController do
         {:error, :bad_request,
          "S3 credentials has to be passed either by request or at application startup as envs"}
 
+      {:error, :overridding_credentials} ->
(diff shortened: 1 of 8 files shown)


lib/jellyfish_web/controllers/component_controller.ex
```diff
@@ -83,6 +83,14 @@ defmodule JellyfishWeb.ComponentController do
         {:error, :bad_request,
          "S3 credentials has to be passed either by request or at application startup as envs"}
 
+      {:error, :overridding_credentials} ->
+        {:error, :bad_request,
+         "Conflicting S3 credentials supplied via environment variables and the REST API. Please provide credentials through only one method"}
```
Rados13, line 88: It's only a suggestion, but from this message, it sounds like the user still can make some changes, e.g., remove `environment_variables`. IMO, the response should mention that this value is already set, and overriding it is prohibited/disabled.

Rados13 (approved):

PR #178, opened 2024-04-10
RTC-512 Add Room.State module and peer disconnected timeout

Rados13:
This PR introduces two changes:
- extracts room state as a separate module and moves part of the logic related to modifying this state to that separate module
- adds support for peer disconnected timeout, after which, if peer doesn't connect, the peer will be automatically removed from the room
- add logs about WebSocket disconnects

[PR to docs](https://github.com/jellyfish-dev/jellyfish-docs/pull/103)
[PR to elixir SDK](https://github.com/jellyfish-dev/elixir_server_sdk/pull/62)
[PR to Python SDK](https://github.com/jellyfish-dev/python-server-sdk/pull/35)


lib/jellyfish_web/api_spec/room.ex
```diff
@@ -46,6 +46,14 @@ defmodule JellyfishWeb.ApiSpec.Room do
           minimum: 1,
           example: 60,
           nullable: true
+        },
+        peerDisconnectedTimeout: %Schema{
+          description:
+            "Duration (in seconds) after which the peer will be removed if it disconnects. If not provided, this feature is disabled.",
```
roznawsk, line 52:
```suggestion
            "Duration (in seconds) after which the peer will be removed if it is disconnected. If not provided, this feature is disabled.",
```

lib/jellyfish_web/api_spec/room.ex
roznawsk, line 1: I find the `peerlessPurgeTimeout` and `peerDisconnectedTimeout` a bit simmilar. How about renaming the `peerlessPurgeTimeout` to `roomIdleTimeout`?

sgfn: The room doesn't have to be idle for the purge to happen, though -- there can be components present...

lib/jellyfish/peer.ex
```diff
@@ -32,7 +39,8 @@ defmodule Jellyfish.Peer do
           socket_pid: pid() | nil,
           engine_endpoint: Membrane.ChildrenSpec.child_definition(),
           tracks: %{Track.id() => Track.t()},
-          metadata: any()
+          metadata: any(),
+          last_time_connected: non_neg_integer() | nil
```
sgfn, line 43: [nitpick] `monotonic_time` returns `integer()`

lib/jellyfish/room.ex
sgfn, line 1: I'm really unsure about the changes introduced here. I see *why* we're choosing to extract the state to a separate module, but as it stands, the implementation is "uneven", the abstraction level fluctuates -- e.g. notifications are sometimes emitted by the functions in state, sometimes by the functions here; calls to State can update a single key in the struct, or do a lot of additional logic as well.

IMO the State module should mainly be considered with, well, keeping track of the room state -- any and all additional tasks, such as emitting telemetry events and notifications, should be performed by Room. However, ensuring this distinction in the code could potentially diminish the benefits we achieved from splitting the state in the first place...

Discussion is welcome

Rados13: I agree that the abstraction level fluctuates.I thought about how to better handle situations like broadcasting events, but it is pretty hard because many events require a lot of information e.g `:track_removed` event the API of the `remove_track` would look terryifing to return all needed information. As I see currently I could move the last two events from Room to State, but I am not sure if this will help a lot.

roznawsk: In my opinion it's a good thing that we moved some of the logic to the `State`. 
It is true however that it is sometimes unclear what actions take place in the `State` and which withing the `Room`.

I think we can clear this up a bit.

First, we can move sending all `Event`s to `State`.
Maybe it would be nice to send them from `Room` . But since all of them but two are emitted from `State` we can just make it a convention.

We can also get rid of any changes to `state` struct within the `Room` and move it to the `State`.

roznawsk (approved):
LGTM 🎸

sgfn (approved):

roznawsk (approved):

PR #180, opened 2024-04-12
Fix ex aws requests

Karolk99:


Rados13 (approved):
As mentioned in our discussion, we can merge this PR as a quick fix, and then we should work on the following:
- adding better integration tests for S3 (e.g by using https://github.com/adobe/S3Mock)
- using only hackney HTTP client in Jellyfish

PR #189, opened 2024-05-10
[RTC-526] Add env vars to configure allowed components

sgfn:
This PR introduces a breaking change: users must now specify which components they wish to use using environment variables.

### DIFF
--- a/lib/jellyfish/component.ex
+++ b/lib/jellyfish/component.ex
@@ -123,7 +123,8 @@ defmodule Jellyfish.Component do
 
   @spec new(component(), map()) :: {:ok, t()} | {:error, term()}
   def new(type, options) do
-    with {:ok, %{endpoint: endpoint, properties: properties}} <-
+    with true <- component_allowed?(type),
+         {:ok, %{endpoint: endpoint, properties: properties}} <-
            type.config(options) do
       {:ok,
        %__MODULE__{
@@ -133,7 +134,17 @@ defmodule Jellyfish.Component do
          properties: properties
        }}
     else
-      {:error, _reason} = error -> error
+      false ->
+        raise("""
+          #{inspect(type)} can only be used if JF_#{type |> to_string!() |> String.upcase()}_USED environmental variable is set to \"true\"
+        """)
+
+      {:error, _reason} = error ->
+        error
     end
+  end
+
+  defp component_allowed?(type) do
+    Application.fetch_env!(:jellyfish, :component_used?)[:"#{to_string!(type)}"]
   end
 end
--- a/lib/jellyfish/config_reader.ex
+++ b/lib/jellyfish/config_reader.ex
@@ -128,6 +128,16 @@ defmodule Jellyfish.ConfigReader do
     end
   end
 
+  def read_components_used() do
+    [
+      hls: read_boolean("JF_HLS_USED") == true,
+      rtsp: read_boolean("JF_RTSP_USED") == true,
+      file: read_boolean("JF_FILE_USED") == true,
+      sip: read_boolean("JF_SIP_USED") == true,
+      recording: read_boolean("JF_RECORDING_USED") == true
+    ]
+  end
+
   def read_sip_config() do
     sip_used? = read_boolean("JF_SIP_USED")
     sip_ip = System.get_env("JF_SIP_IP") || ""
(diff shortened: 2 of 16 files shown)


lib/jellyfish/component.ex
```diff
@@ -133,7 +134,17 @@ defmodule Jellyfish.Component do
          properties: properties
        }}
     else
-      {:error, _reason} = error -> error
+      false ->
+        raise("""
+          #{inspect(type)} can only be used if JF_#{type |> to_string!() |> String.upcase()}_USED environmental variable is set to \"true\"
+        """)
```
sgfn, line 140: This implementation was copied over. However, I'm not sure whether it'd be better to raise here or return an error tuple

roznawsk: I think we should return error tuple here. The motivation is that we can have the same backend with differently configured cloud instances, or the configuration could change for some reason. We don't want to crash the Jellyfish if such thing happens.
You may also change this behavior from wherever you copied it from - if you think it's appropriate.

lib/jellyfish/config_reader.ex
```diff
@@ -128,20 +128,28 @@ defmodule Jellyfish.ConfigReader do
     end
   end
 
+  def read_components_used() do
```
kamciokodzi, line 131: Have you consider injecting this from config file as a list of components instead of a set of true/false key/value? 

What I mean is something like in config.exs or somwewhere:

```elixir
config :jellyfish, 
  components: [
    Jellyfish.Component.SIP,
    Jellyfish.Component.Recording,
    ...
  ]
```

kamciokodzi: In runtime.exs this could be passed as from env which could be a string which we later split and have a ready to go list and is easy to change.

lib/jellyfish/component.ex
sgfn: Changed in WebRTC as well

lib/jellyfish/config_reader.ex
sgfn: Great suggestion. Changed

lib/jellyfish/config_reader.ex
```diff
@@ -128,36 +128,45 @@ defmodule Jellyfish.ConfigReader do
...
 
-  def read_sip_config() do
-    sip_used? = read_boolean("JF_SIP_USED")
+  def read_components_used() do
+    components_used = System.get_env("JF_COMPONENTS_USED") || ""
+
+    components_used
+    |> String.split(" ", trim: true)
+    |> Enum.map(fn type ->
+      case Jellyfish.Component.parse_type(type) do
```
kamciokodzi, line 137: I see that you have already got some "parser" implemented.
For future instead of parsing it "manually" and adding variant to case whenever new component is introduced you could just use `String.to_existing_atom` since every module name is an atom. 

The only possible drawback is that there is a case that one could add non-existing component. In that case you could just check if the component implements some behaviour 😄 But I wouldn't bother with that so much.

lib/jellyfish/room.ex
```diff
@@ -92,7 +94,19 @@ defmodule Jellyfish.Room do
   @spec add_component(id(), Component.component(), map()) ::
           {:ok, Component.t()}
           | :error
-          | {:error, :incompatible_codec | :reached_components_limit_hls}
+          | {:error,
+             {:component_disabled_globally, String.t()}
```
kamciokodzi, line 98: You are using `globally` naming, does it mean something can be enabled locally only?

lib/jellyfish/room.ex
```diff
@@ -92,7 +94,19 @@ defmodule Jellyfish.Room do
...
+             | {:reached_components_limit, String.t()}
+             | :file_does_not_exist
+             | :bad_parameter_framerate_for_audio
+             | :invalid_framerate
+             | :invalid_file_path
+             | :unsupported_file_type
+             | {:missing_parameter, term()}
+             | :missing_s3_credentials
+             | :overriding_credentials
+             | :overriding_path_prefix}
```
kamciokodzi, line 109: Lol a lot of new errors. I assume the previous spec wasn't updated in a long time 😄

roznawsk (approved):

PR #191, opened 2024-05-10
Add logs about errors during adding peer

Rados13:

### DIFF
--- a/lib/fishjam/room_service.ex
+++ b/lib/fishjam/room_service.ex
@@ -135,6 +135,8 @@ defmodule Fishjam.RoomService do
       {:reply, {:ok, room, Fishjam.address()}, state}
     else
       {:error, :room_already_exists} = error ->
+        Logger.warning("Room creation failed, because it already exists")
+
         {:reply, error, state}
 
       reason ->
--- a/lib/fishjam/room_service.ex
+++ b/lib/fishjam/room_service.ex
@@ -24,7 +24,9 @@ defmodule Fishjam.RoomService do
       [{room_pid, _value}] ->
         {:ok, room_pid}
 
-      _not_found ->
+      not_found ->
+        # TODO: Remove this line or debug
+        Logger.warning("Room not found because of: #{inspect(not_found)}")
         {:error, :room_not_found}
     end
   end
--- a/lib/fishjam/room_service.ex
+++ b/lib/fishjam/room_service.ex
@@ -24,9 +24,7 @@ defmodule Fishjam.RoomService do
       [{room_pid, _value}] ->
         {:ok, room_pid}
 
-      not_found ->
-        # TODO: Remove this line or debug
-        Logger.warning("Room not found because of: #{inspect(not_found)}")
+      _not_found ->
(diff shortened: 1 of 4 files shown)


lib/fishjam/room_service.ex
```diff
@@ -133,6 +133,8 @@ defmodule Fishjam.RoomService do
 
   @impl true
   def handle_call({:create_room, config}, _from, state) do
+    Logger.info("Start create room")
+
```
sgfn, line 137: Shouldn't this be `debug` as well?

lib/fishjam/room_service.ex
```diff
@@ -148,7 +150,13 @@ defmodule Fishjam.RoomService do
       {:reply, {:ok, room, Fishjam.address()}, state}
     else
       {:error, :room_already_exists} = error ->
+        Logger.warning("Room creation failed, because it already exists")
+
         {:reply, error, state}
+
+      reason ->
+        Logger.warning("Room creation failed with reason: #{inspect(reason)}")
+        {:reply, :room_doesnt_start, state}
```
sgfn, line 159: 1. Shouldn't this be `{:error, :room_doesnt_start}`?
2. This needs to be handled in `RoomController` as well

lib/fishjam_web/controllers/peer_controller.ex
```diff
@@ -86,19 +89,30 @@ defmodule FishjamWeb.PeerController do
       |> render("show.json", assigns)
     else
       :error ->
-        {:error, :bad_request, "Invalid request body structure"}
+        msg = "Invalid request body structure"
+        Logger.warning(msg)
```
sgfn, line 93: Shouldn't we specify the action that failed? I.e. `"Unable to add peer to room #{room_id}: " <> msg`?

lib/fishjam_web/controllers/room_controller.ex
```diff
@@ -107,6 +107,10 @@ defmodule FishjamWeb.RoomController do
         room_id = Map.get(params, "roomId")
         {:error, :bad_request, "Cannot add room with id \"#{room_id}\" - room already exists"}
 
+      {:error, :room_doesnt_start} ->
+        room_id = Map.get(params, "roomId")
+        {:error, :bad_request, "Cannot add room with id \"#{room_id}\" - adding failed"}
```
sgfn, line 112: How about "unexpected error" instead? I feel like we're repeating ourselves with "adding failed"

sgfn (approved):

lib/fishjam_web/controllers/room_controller.ex
```diff
@@ -74,6 +75,8 @@ defmodule FishjamWeb.RoomController do
   end
 
   def create(conn, params) do
+    Logger.debug("Start creating room")
```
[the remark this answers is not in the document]
Rados13: With creating room, is that difference that the log from `room_controller` will log on a node to which the request comes, and the log from `room_service` will log on the node where this room is created? That's why I see benefits in having these two logs, but maybe the messages should be different.

PR #197, opened 2024-05-16
Configure log level

Rados13:
 
[ PR to FJ docs](https://github.com/fishjam-dev/fishjam-docs/pull/111)

### DIFF
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -26,5 +26,3 @@ config :phoenix, :stacktrace_depth, 20
 
 # Initialize plugs at runtime for faster development compilation
 config :phoenix, :plug_init_mode, :runtime
-
-config :logger, level: :info
--- a/lib/jellyfish/component/hls/manager.ex
+++ b/lib/jellyfish/component/hls/manager.ex
@@ -76,11 +76,7 @@ defmodule Fishjam.Component.HLS.Manager do
 
     broadcast_notification(result, room_id)
 
-    if is_tuple(result) do
-      Logger.info("Finished uploading to s3 with result: #{inspect(result)}, room: #{room_id}")
-    else
-      Logger.info("Finished uploading to s3 with result: #{result}, room: #{room_id}")
-    end
+    Logger.info("Finished uploading to s3 with result: #{inspect(result)}, room: #{room_id}")
   end
 
   defp upload_file_to_s3(content, s3_path, opts, config, credentials) do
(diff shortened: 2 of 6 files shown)


lib/jellyfish/component/hls/manager.ex
```diff
@@ -75,7 +75,12 @@ defmodule Fishjam.Component.HLS.Manager do
       end)
 
     broadcast_notification(result, room_id)
-    Logger.info("Finished uploading to s3 with result: #{result}, room: #{room_id}")
+
+    if is_tuple(result) do
+      Logger.info("Finished uploading to s3 with result: #{inspect(result)}, room: #{room_id}")
```
kamciokodzi, line 80: I think this if is not necessary, you could just use this one in all cases.

lib/jellyfish/config_reader.ex
```diff
@@ -246,6 +246,23 @@ defmodule Fishjam.ConfigReader do
     get_env("FJ_GIT_COMMIT", "dev")
   end
 
+  def read_logger_level() do
+    log_level = System.get_env("FJ_LOG_LEVEL", "info")
```
sgfn, line 250: I wonder whether we should handle `JF_LOG_LEVEL` as well. I know this env var didn't exist prior to the rebranding, but I think it's better to be consistent in this matter. In a few versions we'll drop the `JF_` support anyway

```suggestion
    log_level = get_env("FJ_LOG_LEVEL", "info")
```

sgfn (approved):

config/dev.exs
```diff
@@ -26,5 +26,3 @@ config :phoenix, :stacktrace_depth, 20
 
 # Initialize plugs at runtime for faster development compilation
 config :phoenix, :plug_init_mode, :runtime
-
-config :logger, level: :info
```
kamciokodzi, line 30: Why? Do we want to include `debug` logs by default as well?

lib/jellyfish/config_reader.ex
```diff
@@ -246,6 +246,23 @@ defmodule Fishjam.ConfigReader do
...
+
+    log_levels_string = ["info", "debug", "warning", "error"]
+
+    if log_level in log_levels_string do
+      String.to_atom(log_level)
```
kamciokodzi, line 255: I would consider `String.to_existing_atom/1`

config/dev.exs
mickel8: Now, log level is configured in runtime.exs where info is the default value

mickel8 (approved):
Can we merge this after https://github.com/fishjam-dev/fishjam/pull/199?

config/dev.exs
kamciokodzi: Ah yes my mistake, thought it was in prod not runtime 😅

kamciokodzi (approved):

PR #200, opened 2024-05-17
Feat/introduce ADRs

kamciokodzi:


docs/adrs/002-change-release-strategy.md
```diff
@@ -0,0 +1,78 @@
...
+This is an awful user experience. Another issue is the fact that any ongoing recordings are lost.
+
+## Decision Drivers
+
+* N o downtime while release happens
+* 0 lost recordings during release
+
+## Considered Options
+
+* Rolling update with eviction
```
mickel8, line 21: Should we write more on the API? E.g. whether we want to add a new REST endpoint?

mickel8 (commented):
Awesome, I added a few minor comments :+1: Also, what are we going to do with the old design docs? Maybe create a new directory for them called `old`?

docs/adrs/002-change-release-strategy.md
```diff
@@ -0,0 +1,78 @@
...
+External process will trigger fishjam shutdown process which is gonna mark application as one that no longer accepts the connections (new rooms etc.). Once every room is closed and all of the recordings are computed we are going to mark fishjam as ready to shutdown. This application instance should also be effectively removed from any external load balancing (k8s etc.) while it should still allow erlang clustering. Orchestrator process should query the fishjam till it's ready to shutdown and sends sigterm signal.
+
+This is applicable one by one for every instance in cluster, although it may take some time to release a new version and we may have 2 different versions on the cluster at the same time, we are accepting that tradeoff. We may consider `force` option to release a version no matter the state of fishjam (which may result in downtime).
+
+## Pros and Cons of the Options
+
+### Rolling update with eviction
+
+1. External orchestrator process triggers new release
+2. Starts new instance of fishjam with new version
```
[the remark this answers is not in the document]
kamciokodzi: Then we will have (N-1) instances in cluster available to handle the same amount of traffic.

docs/adrs/002-change-release-strategy.md
```diff
@@ -0,0 +1,78 @@
...
+5. Once peers/rooms/streams are migrated app is ready for sigterm
+6. External orchestrator process kills fishjam instance
+7. Repeat for rest of the remaining instances
+
+* Good, because we will eliminate downtimes with releases
+* Good, because we won't lose any recordings
+* Good, because it happens fast, we don't have to wait for rooms to close/streams to end
+* Good, because we will have 2 different versions on cluster for a small amount of time
+* Neutral, because we require some external work from the orchestrator
+* Bad, because we don't have a clue how to handle the active peer migration to new instance right now
```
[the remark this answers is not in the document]
kamciokodzi: Haha! Yeah that's right I will correct this.

docs/adrs/002-change-release-strategy.md
```diff
@@ -0,0 +1,78 @@
...
+This is applicable one by one for every instance in cluster, although it may take some time to release a new version and we may have 2 different versions on the cluster at the same time, we are accepting that tradeoff. We may consider `force` option to release a version no matter the state of fishjam (which may result in downtime).
+
+## Pros and Cons of the Options
+
+### Rolling update with eviction
+
+1. External orchestrator process triggers new release
+2. Starts new instance of fishjam with new version
+3. Fishjam shutdown process start
+4. Stop receiving traffic to that instance
```
Rados13, line 44: Shouldn't we specify which traffic we won't receive? If I remember correctly, we mentioned only requests about creating a new room; all other traffic should be allowed.

kamciokodzi:
> Awesome, I added a few minor comments 👍 Also, what are we going to do with the old design docs? Maybe create a new directory for them called `old`?

Well I assumed [here](https://github.com/fishjam-dev/fishjam/pull/200/files#diff-6cef50a16144f58dd82dcc4940ec80886db3bb7a4c3df1172b92bfa4874d4ff0R42) that we won't write ADRs for old decisions. I wouldn't also call them `old` since they are up to date with the current architecture. I don't mind having them laying around in the `docs` folder, but I'm open to suggestions.

docs/adrs/002-change-release-strategy.md
Rados13, line 1: Shouldn't we specify in this document what has to be implemented/modified in the chosen approach?

docs/adrs/002-change-release-strategy.md
```diff
@@ -0,0 +1,78 @@
+---
+status: proposed
+date: 17-05-2024
+deciders: Kamil Kołodziej, Radosław Szuma, Jakub Pisarek
+informed: Fishjam Team, Cloud Team
+---
+# Change release strategy to eliminate downtimes
```
PiotrWodecki, line 7: **nit**: downtime is uncountable
```suggestion
# Change release strategy to eliminate downtime
```

docs/adrs/002-change-release-strategy.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Context and Problem Statement
+
+In the current situation when we do a release we are shuting down the server and every room with that.
+This is an awful user experience. Another issue is the fact that any ongoing recordings are lost.
+
+## Decision Drivers
+
+* N o downtime while release happens
```
PiotrWodecki, line 16: **nit**: typo
```suggestion
* No downtime while release happens
```

docs/adrs/002-change-release-strategy.md
```diff
@@ -0,0 +1,78 @@
...
+## Pros and Cons of the Options
+
+### Rolling update with eviction
+
+1. External orchestrator process triggers new release
+2. Starts new instance of fishjam with new version
+3. Fishjam shutdown process start
+4. Stop receiving traffic to that instance
+5. Wait till the last room is closed and all recordings are completed
+6. Finish fishjam shutdown process - ready for sigterm
```
PiotrWodecki, line 46: **strong suggestion**: In most systems/contexts (docker, k8s, bare-metal) `SIGTERM` should be the initial signal sent to the application when requesting a graceful shutdown. I strongly suggest outlining this fact here.

I think we can easily call this a contract between the Cloud and Fishjam teams.

docs/adrs/002-change-release-strategy.md
PiotrWodecki, line 44: **issue**: It should be specified **who** is responsible for directing the traffic to the right instance, especially with future Cluster API changes.

docs/adrs/002-change-release-strategy.md
```diff
@@ -0,0 +1,78 @@
...
+7. External orchestrator process kills fishjam instance
```
PiotrWodecki, line 47: **issue**: When the instance receives a `SIGTERM` - after cleaning up, it should quit by itself, not on an additional orchestrator request.

docs/adrs/002-change-release-strategy.md
```diff
@@ -0,0 +1,78 @@
...
+8. Repeat for rest of the remaining instances
```
PiotrWodecki, line 48: **suggestion**: This step depends on the orchestrator's configuration and is more of a suggestion than a strict requirement. I'd focus on signal handling and clustering behavior in this ADR, not the exact external behavior.

docs/adrs/002-change-release-strategy.md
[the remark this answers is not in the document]
PiotrWodecki: I don't agree. ;P
The cost of a solution is definitely a drawback that is considered when making a decision. Because ADR is quite literally, a record of a decision, it should contain cost doubts and rationale for why we've made a specific choice at this time.

docs/adrs/002-change-release-strategy.md
```diff
@@ -0,0 +1,78 @@
...
+* Good, because we won't lose any recordings
+* Good, because it happens fast, we don't have to wait for rooms to close/streams to end
+* Good, because we will have 2 different versions on cluster for a small amount of time
+* Neutral, because we require some external work from the orchestrator
+* Bad, because we don't have a clue how to handle the active peer migration to new instance right now
+* Bad, because we don't know how to handle the meetings with recording enabled during the migration
+
+## More Information
+
+This is decision which is heavily dependent on the Cloud Team and may be change soon to meet their requirements.
```
PiotrWodecki, line 78: **nit**: grammar
```suggestion
This decision is heavily dependent on the Cloud Team and may be changed soon to meet their requirements.
```

docs/adrs/002-change-release-strategy.md
```diff
@@ -0,0 +1,78 @@
...
+* Active room migration
+
+## Decision Outcome
+
+Chosen option: "Rolling update with eviction", because
+while second option sounds great, we acknowledge that solution is not trivial and we want to fix downtimes ASAP. First solution is a good starting point to remove existing problems and allow us to move forward.
+
+### Consequences
+
+Every new release will need to trigger fishjam process which will handle the shutdown. What that effectively means is that we are leaving the responsibility of triggering release to external orchestration tool (what is usually the case) and to monitor if the application exited correctly.
```
PiotrWodecki, line 31: **issue**: So, release != deployment or upgrade. I suggest generally replacing this word everywhere in this ADR with something else. IMHO it makes it harder to understand, as this has not much in common with the Fishjam release cycle.

```suggestion
Every new deployment will need to trigger fishjam process which will handle the shutdown. What that effectively means is that we are leaving the responsibility of triggering release to external orchestration tool (what is usually the case) and to monitor if the application exited correctly.
```

docs/adrs/001-start-using-adrs.md
```diff
@@ -0,0 +1,42 @@
+---
+status: decided
```
PiotrWodecki, line 2: **nit**
```suggestion
status: accepted
```

docs/adrs/001-start-using-adrs.md
```diff
@@ -0,0 +1,42 @@
...
+date: 17-05-2024
```
PiotrWodecki, line 3: nit: YYYY-MM-DD seems a more common and sensible format

docs/adrs/001-start-using-adrs.md
```diff
@@ -0,0 +1,42 @@
...
+deciders: Kamil Kołodziej
```
PiotrWodecki, line 4: **issue**: ADR format introduction should definitely be decided by a majority of the Fishjam team.

docs/adrs/002-change-release-strategy.md
```diff
@@ -0,0 +1,78 @@
...
+3. Fishjam starts migrating rooms to new instance
+4. Somehow handles the recordings (?)
+5. Once peers/rooms/streams are migrated app is ready for sigterm
+6. External orchestrator process kills fishjam instance
+7. Repeat for rest of the remaining instances
+
+* Good, because we will eliminate downtimes with releases
+* Good, because we won't lose any recordings
+* Good, because it happens fast, we don't have to wait for rooms to close/streams to end
+* Good, because we will have 2 different versions on cluster for a small amount of time
```
PiotrWodecki, line 71: **nit**: grammar
```suggestion
* Good, because we will have 2 different versions on cluster for a short amount of time
```

docs/adrs/001-start-using-adrs.md
```diff
@@ -0,0 +1,42 @@
...
+
+* One place to keep all the decisions made
+* Easy to read and write
+* Close to the code
+
+## Considered Options
+
+* ADRs inside the repo
+* Google Docs
+* Confluence
```
p1003, line 23: **question**: Shouldn't `Google Docs` and `Confluence` be at least shortly described somewhere in this ADR?

docs/adrs/002-change-release-strategy.md
pgronkievitz: note that grace period between sigterm and sigkill is brief for most orchestrators iirc like 30 seconds by default) and we may want to handle it properly

docs/adrs/002-change-release-strategy.md
sgfn: Seconding this, in favour of 'deployment' as it can refer to both an 'upgrade' and a 'downgrade'/'downpatch'

docs/adrs/002-change-release-strategy.md
```diff
@@ -0,0 +1,78 @@
...
+## Decision Outcome
+
+Chosen option: "Rolling update with eviction", because
+while second option sounds great, we acknowledge that solution is not trivial and we want to fix downtimes ASAP. First solution is a good starting point to remove existing problems and allow us to move forward.
+
+### Consequences
+
+Every new release will need to trigger fishjam process which will handle the shutdown. What that effectively means is that we are leaving the responsibility of triggering release to external orchestration tool (what is usually the case) and to monitor if the application exited correctly.
+
+External process will trigger fishjam shutdown process which is gonna mark application as one that no longer accepts the connections (new rooms etc.). Once every room is closed and all of the recordings are computed we are going to mark fishjam as ready to shutdown. This application instance should also be effectively removed from any external load balancing (k8s etc.) while it should still allow erlang clustering. Orchestrator process should query the fishjam till it's ready to shutdown and sends sigterm signal.
```
sgfn, line 33: I wouldn't mention 'no longer accepting the connections', this can paint the wrong picture and isn't entirely in line with what we decided (the fishjam that's shutting down should still be able to relay any and all requests to others in the cluster).

docs/adrs/002-change-release-strategy.md
PiotrWodecki: All considered platforms support an infinite (or long enough) grace period for it not to be a problem. The user or the cloud team will just have to handle it on their side.

docs/adrs/002-change-release-strategy.md
sgfn: Right now, business logic (BE); after the API changes, every FJ in the cluster will be able to route the requests.

I don't think this is the right place to specify that, I think a separate ADR about the new clustering behaviour should (and will) be created; then we can simply link it here. Alternatively, we could add a single sentence like the first one in this comment

PiotrWodecki: That would be good enough. I'm simply wondering whether it is supposed to be implemented as a load-balancer configuration or will it be done automagically by Fishjam.

docs/adrs/001-start-using-adrs.md
kamciokodzi: Well maybe in something more serious (like the one below) I'm all for describing the options but here I'm just not feeling the need. One could deduct from the Decision outcome why we don't choose the other two.

docs/adrs/002-change-release-strategy.md
```diff
@@ -0,0 +1,78 @@
...
+3. Fishjam shutdown process start
+4. Stop receiving traffic to that instance
+5. Wait till the last room is closed and all recordings are completed
+6. Finish fishjam shutdown process - ready for sigterm
+7. External orchestrator process kills fishjam instance
+8. Repeat for rest of the remaining instances
+
+* Good, because we will eliminate downtimes with releases
+* Good, because we won't lose any recordings
+* Neutral, because we require some external work from the orchestrator
```
bblaszkow06: This is "bad", not "neutral" and should be removed by considering the following (some things were already mentioned in the other comments):
* The ADR should cover how to pass the info about shutting down. Remember that REST endpoint would have to be somehow secured to be available only to some orchestrator, so using SIGTERM seems like an easier option
* FJ should not expect orchestrator to check shutdown readiness after sending "soft-shutdown" request and should shutdown automatically

kamciokodzi: already resolved in [74ec204](https://github.com/fishjam-dev/fishjam/pull/200/commits/74ec20407a0aded6e052828b3fd5d1d0c4f1a898)

docs/adrs/002-change-release-strategy.md
kamciokodzi: This is a decision in what direction we want to go, what are the architecture drivers, consequences, other options etc.

 What needs to be done is a topic which should be discussed within the team and should comply with the outcome of this document.

PiotrWodecki (approved):
LGTM - good work!

docs/adrs/002-use-rolling-update-with-eviction.md
```diff
@@ -0,0 +1,82 @@
...
+### Consequences
+
+Every new deployment will need to trigger fishjam process which will handle the shutdown.
+What that effectively means is that we are leaving the responsibility of triggering deployment to external orchestration tool (what is usually the case) but we handle the process inside the app.
+
+External process will trigger fishjam shutdown process by a SIGTERM.
+This is gonna mark fishjam as one that no longer accepts new rooms.
+Once every room is closed and all of the recordings are computed we are going to shutdown that instance of fishjam.
+
+This is applicable one by one for every instance in cluster, although it may take some time to deployment a new version and we may have 2 different versions on the cluster at the same time, we are accepting that tradeoff.
```
mickel8, line 38: As we discussed, having two different versions in a cluster might be a problem as those versions might not be backward and forward compatible. We can take a look what rabbit does https://www.rabbitmq.com/docs/upgrade

kamciokodzi: Based on the Rabbit approach they allow rolling updates in some scenarios but when there is some breaking change (version incompatibility) they require you to do full stop or blue-green deployment. 

Their approach to blue-green deployment is to completely create a second cluster and slowly move traffic to the new one while draining the queues from the old one. This is very interesting approach but I don't find it necessary to dig it into now. I would propose moving forward with rolling updates approach and full stop. If that's gonna happen often then I suggest we come back to this discussion.

PR #206, opened 2024-05-28
feat: introduce room_id generator

kamciokodzi:
This PR introduces:
- Room ID generator based on the node name where the room was created
- Determining which node owns given room
- Replaced `rpc` with `erpc`
- Introduce primitive feature flag module

[Docs](https://github.com/fishjam-dev/fishjam-docs/pull/119)


lib/fishjam/room/config.ex
```diff
@@ -30,7 +34,11 @@ defmodule Fishjam.Room.Config do
 
   @spec from_params(map()) :: {:ok, __MODULE__.t()} | {:error, atom()}
   def from_params(params) do
-    room_id = Map.get(params, "roomId")
+    room_id =
+      if FeatureFlags.custom_room_name_disabled?(),
+        do: ID.generate(),
+        else: Map.get(params, "roomId")
```
Rados13, line 40: I think that logic regarding `room_id` should be in the same place; currently, we separately have this, and separately, we have to parse room_id.

lib/fishjam/room/id.ex
Rados13, line 1: I think we should add unit tests for that module.

lib/fishjam/rpc_client.ex
```diff
@@ -0,0 +1,50 @@
...
+  defp handle_result(result) when is_list(result) do
+    result
+    |> Enum.reduce([], fn {status, res}, acc ->
+      if status == :ok do
+        [res | acc]
+      else
+        Logger.warning("RPC multicall to one of the nodes failed because of: #{inspect(res)}")
+        acc
+      end
+    end)
```
Rados13, line 46:
```suggestion
    |> Enum.reduce([], fn 
    {:ok, res}, acc ->
        [res | acc]
    {status, res}, acc -> 
        Logger.warning("RPC multicall to one of the nodes failed with status #{status} because of: #{inspect(res)}")
        acc
      end
    end)
```

lib/fishjam/rpc_client.ex
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  @doc """
+  Multicall to all nodes in cluster.
+  It filters out any errors or exceptions from return so you may end up with empty list.
+  """
+  @spec multicall(module(), atom(), term(), timeout()) :: list(term)
+  def multicall(module, function, args, timeout \\ :infinity) do
+    :erpc.multicall(nodes(), module, function, args, timeout)
+    |> handle_result()
```
Rados13, line 34: What do you think [about this point](https://github.com/christopheradams/elixir_style_guide?tab=readme-ov-file#avoid-single-pipelines) of elixir style guide?

kamciokodzi: Ah that's actually good catch! Usually I use those checks in credo which should detect this situation:

1. https://github.com/rrrene/credo/blob/0058645812e29ac63a640905a2ad4958fc7bafd9/lib/credo/check/readability/single_pipe.ex#L5

2. https://github.com/rrrene/credo/blob/0058645812e29ac63a640905a2ad4958fc7bafd9/lib/credo/check/refactor/pipe_chain_start.ex#L5

kamciokodzi: Although first one is something that I find myself controversial.
I really like pipe and even use it in just once 😄

lib/fishjam/room/id.ex
```diff
@@ -0,0 +1,86 @@
...
+    end
+  end
+
+  defp decode_node_name(room_id) do
+    room_id
+    |> String.split("-")
+    |> Enum.take(-1)
+    |> Enum.at(0)
+    |> Base.decode16!(case: :lower)
+    |> String.to_atom()
```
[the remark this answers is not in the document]
kamciokodzi: Actually it should be `String.to_existing_atom` my bad - while testing it on PoC I had to replace it with `to_atom` cuz it was just simple app. Copy paste mistake.

lib/fishjam/room/id.ex
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+  defp encoded_node_name do
+    Node.self()
+    |> Atom.to_string()
+    |> Base.encode16(case: :lower)
```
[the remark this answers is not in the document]
kamciokodzi: Generally I agree we can improve current solution with the proper hashing.

The only downside I see is that we won't be able to determine the nodename from the room_id (we always need to hash nodes to get the matching) - possibly closing the door for some debugging, not sure if that's relevant right now.

lib/fishjam/feature_flags.ex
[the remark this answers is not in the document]
kamciokodzi: This should be added in another PR.

Rados13 (approved):

sgfn (approved):

PR #211, opened 2024-06-11
[RTC-547] Redirect websockets messages

Rados13:


lib/fishjam/room.ex
```diff
@@ -401,14 +412,34 @@ defmodule Fishjam.Room do
   @impl true
   def handle_info(%EndpointMessage{endpoint_id: to, message: {:media_event, data}}, state) do
     with {:ok, peer} <- State.fetch_peer(state, to),
-         socket_pid when is_pid(socket_pid) <- Map.get(peer, :socket_pid) do
-      send(socket_pid, {:media_event, data})
+         socket_pid when is_pid(socket_pid) <- Map.get(peer, :socket_pid),
+         node_name when not is_nil(node_name) <- Map.get(peer, :node_name),
```
kamciokodzi, line 416: Shouldn't happen in any scenario, media events comes after establishing connection. I would prefer to crash when this happens than `handle` it 🤔

lib/fishjam/room.ex
```diff
@@ -401,14 +412,34 @@ defmodule Fishjam.Room do
...
+         {:ok, rpc_result} <-
```
kamciokodzi, line 417: This could also be `{:ok, :ok}`. Is there any thought behind handling it in the way you do?

lib/fishjam/room.ex
```diff
@@ -401,14 +412,34 @@ defmodule Fishjam.Room do
...
           "Received Media Event from RTC Engine to peer #{inspect(to)} without established signaling connection"
         )
 
+      :error_rpc ->
+        Logger.warning(
+          "Received Media Event from RTC Engine to peer #{inspect(to)} which was connected to different node and rpc to that node failed."
+        )
+
+        {:ok, peer} = State.fetch_peer(state, to)
+        disconnect_peer(peer.socket_pid, self(), :peer_socket_not_exists)
```
kamciokodzi, line 441:
```suggestion
        disconnect_peer(peer.socket_pid, self(), :error_rpc)
```

or something like that

test/fishjam/cluster/load_balancing_test.exs
```diff
@@ -53,11 +62,98 @@ defmodule Fishjam.Cluster.LoadBalancingTest do
...
+         room_id: room_id,
+         token: token,
+         room_instance: fishjam_instance,
+         other_instance: other_node,
+         peer_id: peer_id
+       }}
+    end
+
+    @tag timeout: @max_test_duration
+    test "delete room on other node", state do
```
kamciokodzi, line 93: I think this one and one on `99` tests HTTP API not the WS since you are setting up ws and then make a request to controller not a ws message.

Rados13: It's testing that the peer socket is correctly closed on another node when the room is another node, and it is deleted. (So it is testing the logic forwarding information about deleting peer/room from one node to another).

.circleci/config.yml
```diff
@@ -30,6 +30,7 @@ jobs:
       - checkout
       - run: mix deps.get
       - run: mix coveralls.json --warnings-as-errors
+      - run: FJ_FEATURE_FLAG_CUSTOM_ROOM_NAME_DISABLED=true mix coveralls.json --warnings-as-errors
```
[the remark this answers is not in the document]
Rados13: The problem with adding tests with a flag enabled/disabled is that more code is required to run the same tests. Also, you must remember that you will always have to write two tests. Metaprogramming could be an option, but it has some disadvantages when generating tests, e.g., it is hard to call one specific test if it is generated. IMO, this approach is the easiest one, and it gives the best prevention against introducing bugs from misconfiguration. We could run it in separate actions to make it faster.

lib/fishjam/room.ex
```diff
@@ -139,6 +141,12 @@ defmodule Fishjam.Room do
     GenServer.cast(registry_id(room_id), {:media_event, peer_id, event})
   end
 
+  @spec disconnect_peer(socket_pid :: pid(), room_pid :: pid(), reason :: atom()) :: :ok
+  def disconnect_peer(socket_pid, room_pid, reason) do
+    send(room_pid, {:DOWN, :any_ref, socket_pid, reason})
```
sgfn, line 146: 1. Should be `{:DOWN, ref, :process, pid, reason}`...
2. ...but impersonating a `:DOWN` tuple feels... wrong

sgfn (commented):
It is unfortunate, but whoever merges his PR first will leave the other person with lots of conflicts. We can talk this issue through in the next few days

lib/fishjam/room.ex
Rados13: The main thought was that matching in else on `{:ok, :error_rpc}` was strange for me and I tried to avoid that.

test/fishjam/room/id_test.exs
```diff
@@ -12,10 +12,30 @@ defmodule Fishjam.Room.IDTest do
     end
 
     test "returns error if node is not detected in cluster" do
+      old_values = Application.get_env(:fishjam, :feature_flags)
```
kamciokodzi, line 15: can we move this to setup?

lib/fishjam_web/peer_socket.ex
```diff
@@ -22,17 +25,20 @@ defmodule FishjamWeb.PeerSocket do
...
   @impl true
   def handle_in({encoded_message, [opcode: :binary]}, %{authenticated?: false} = state) do
     case PeerMessage.decode(encoded_message) do
       %PeerMessage{content: {:auth_request, %AuthRequest{token: token}}} ->
         with {:ok, %{peer_id: peer_id, room_id: room_id}} <- PeerToken.verify(token),
-             {:ok, room_pid} <- RoomService.find_room(room_id),
-             :ok <- Room.set_peer_connected(room_id, peer_id),
+             {:ok, node_name} <- get_node_name(room_id),
+             args = [room_id, peer_id, Node.self(), self()],
+             {:ok, connect_result} <-
```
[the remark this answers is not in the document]
Rados13: Nope, we can't because from this call, we can receive, e.g.: `{:error, room_not_found},` so in the else branch, we would have to added another branch to handle `{:ok, {:error, reason}},` which we could add but IMO it would look bad.

kamciokodzi (approved):
Let's fix the conflicts and merge it 😄 🎉

sgfn (approved):
LGTM 🥇

PR #212, opened 2024-06-12
[RTC-543] Request routing first iteration

sgfn:


lib/fishjam/cluster/room_service.ex
```diff
@@ -0,0 +1,78 @@
...
+          %{forwarded_tracks_number: forwarded_tracks1},
+          %{forwarded_tracks_number: forwarded_tracks2} ->
+            forwarded_tracks1 < forwarded_tracks2
+        end
+      )
+
+    min_node
+  end
+
+  defp route_request(room_id, fun, args) do
```
kamciokodzi, line 67: I wonder if this is recurring function maybe we can move it to RPC module 🤔

kamciokodzi: Although I see that here `lib/fishjam/component/hls/cluster/request_handler.ex` we have a little bit different implementation.

lib/fishjam/feature_flags.ex
```diff
@@ -5,11 +5,22 @@ defmodule Fishjam.FeatureFlags do
...
-  Removal: Once we move on to generated room_ids permanently.
+  Removal: Once we move on to generated room_ids and cluster-wide request routing permanently.
   """
-  def custom_room_name_disabled?,
-    do: Application.get_env(:fishjam, :feature_flags)[:custom_room_name_disabled]
+  def request_routing_enabled?(),
+    do: Application.get_env(:fishjam, :feature_flags)[:request_routing_enabled?]
+
+  @doc "Info about currently enabled feature flags"
+  def info() do
```
kamciokodzi, line 18: Functions without args should be written without parentheses.

https://github.com/christopheradams/elixir_style_guide?tab=readme-ov-file#fun-def-parentheses

lib/fishjam/room/id.ex
```diff
@@ -8,20 +8,17 @@ defmodule Fishjam.Room.ID do
...
   @spec determine_node(id()) ::
-          {:ok, node()} | {:error, :invalid_room_id} | {:error, :invalid_node}
+          {:ok, node()} | {:error, :invalid_room_id | :node_not_found}
   def determine_node(room_id) do
     with {:ok, room_id} <- validate_room_id(room_id),
-         node_name <- decode_node_name(room_id),
+         {:ok, node_name} <- decode_node_name(room_id),
          true <- node_present_in_cluster?(node_name) do
-      {:ok, node_name}
+      {:ok, String.to_atom(node_name)}
```
kamciokodzi, line 18: `String.to_existing_atom` to prevent creating new atoms. Node name is a registered atom so it shouldn't ever fail.

lib/fishjam_web/controllers/component_controller.ex
```diff
@@ -76,12 +79,18 @@ defmodule FishjamWeb.ComponentController do
       :error ->
         {:error, :bad_request, "Invalid request body structure"}
 
-      {:error, :room_not_found} ->
-        {:error, :not_found, "Room #{room_id} does not exist"}
-
       {:error, :invalid_type} ->
         {:error, :bad_request, "Invalid component type"}
 
+      {:error, :invalid_room_id} ->
```
kamciokodzi, line 85: I feel like we could have `fallback_controller` where we would catch all those `generic` errors with invalid_room_id rpc_failed etc.

kamciokodzi: And it could be used in all controllers which use RPC underneath.

lib/fishjam_web/controllers/hls_content_controller.ex
```diff
@@ -97,11 +104,27 @@ defmodule FishjamWeb.HLSContentController do
...
     end
   end
+
+  defp send_redirect_response(conn, fishjam_address) do
+    location =
+      "#{conn.scheme}://#{fishjam_address}"
+      |> URI.parse()
+      |> Map.put(:path, conn.request_path)
+      |> URI.to_string()
+      |> IO.inspect(label: :redirect_location)
```
kamciokodzi, line 124: inspect detected 👀

lib/fishjam/cluster/room_service.ex
sgfn: Yeah I didn't really see an easy way to reuse the same parts of the function

lib/fishjam/local/room_service.ex
```diff
@@ -0,0 +1,235 @@
...
+    :ok = Phoenix.PubSub.subscribe(Fishjam.PubSub, "fishjams")
+    {:noreply, state}
+  end
+
+  @impl true
+  def handle_call({:create_room, config}, _from, state) do
+    Logger.debug("Creating a new room")
+
+    # Re-generate the room ID to ensure it has the correct node name part
+    {:ok, room_id} = ID.generate(config.room_id)
```
kamciokodzi, line 115: Oh shiet that was a huge bug lol. I feel like instead of having it there the logic behind it should be somewhere else since now it's sooo bad to have it in config but the data is not consistent etc etc.

Something to definitely refactor soon after this PR is merged.

lib/fishjam_web/controllers/component_controller.ex
```diff
@@ -129,8 +132,11 @@ defmodule FishjamWeb.ComponentController do
          :ok <- Room.remove_component(room_id, id) do
       send_resp(conn, :no_content, "")
     else
-      {:error, :room_not_found} -> {:error, :not_found, "Room #{room_id} does not exist"}
-      {:error, :component_not_found} -> {:error, :not_found, "Component #{id} does not exist"}
+      {:error, :component_not_found} ->
+        {:error, :not_found, "Component #{id} does not exist"}
+
+      other ->
```
kamciokodzi, line 138: If I remember correctly there is no need to have this `other -> other` fallback should automatically catch those.

test/fishjam/cluster/api_test.exs
```diff
@@ -0,0 +1,248 @@
...
+          for room <- rooms, do: HTTPoison.delete("http://#{node}/room/#{room["id"]}", @headers)
+        end
+      end
+    end)
+
+    {:ok, %{nodes: @nodes}}
+  end
+
+  @tag timeout: @max_test_duration
+  test "load-balancing and listing rooms in a cluster", %{nodes: [node1, node2]} do
```
kamciokodzi, line 37: To be honest I feel like a lot is tested and happening here in that test.
What do you think maybe of dividing it into smaller tests some scenarios or alternatively comment before the next chunk of code and asserts whats gonna happen. 
Right now it's a huge bulk of asserts and you really need to read the whole block to see what's going on. 

Maybe some scenarios for separate tests
- One room added, 1 on one node 0 on the other one
- Two rooms added, they are split between nodes
- Two rooms, we delete one room 

So basically the same as you have here but split between tests so we have easier time reading it even if we add a few lines of code.

test/fishjam/cluster/api_test.exs
```diff
@@ -0,0 +1,248 @@
...
+
+    assert_room_count_in_cluster(room_node, 0)
+    assert_room_count_in_cluster(other_node, 0)
+  end
+
+  @tag timeout: @max_test_duration
+  test "request routing + explicit forwarding of HLS retrieve content requests", %{
+    nodes: [node1, node2]
+  } do
+    response_body1 = add_room(node1)
```
kamciokodzi, line 113: Maybe modify `add_room` so it could return 
`{room_data, room_node}  = add_room(node1)`

this way we won't need to call `get_fishajm_address` in every test

also I'm wondering if it wouldn't be better to already decode json inside those http requests instead of doing it in tests

lib/fishjam/local/room_service.ex
sgfn: Yeah, we missed that the node_name was taken from the FJ which received the request...
I'll mark this with a `FIXME`

Sidenote, there are also other clear opportunities for improvement -- most notably, right now we perform the RPC twice whenever doing something with an existing room (once in `RoomService`, once in `Room`). It's totally possible to do this in a single RPC -- however, since it's the first iteration, I didn't really want to make changes this big.

lib/fishjam_web/controllers/component_controller.ex
sgfn: Not in this case :c
```

  1) test delete peer deletes peer from not exisiting room (FishjamWeb.PeerControllerTest)
     test/fishjam_web/controllers/peer_controller_test.exs:132
     ** (WithClauseError) no with clause matching: {:error, :room_not_found}
     code: conn = delete(conn, ~p"/room/#{room_id}/peer/#{peer_id}")
     stacktrace:
       (fishjam 0.6.3) lib/fishjam_web/controllers/peer_controller.ex:120: FishjamWeb.PeerController.delete/2
       (fishjam 0.6.3) lib/fishjam_web/controllers/peer_controller.ex:1: FishjamWeb.PeerController.action/2
       (fishjam 0.6.3) lib/fishjam_web/controllers/peer_controller.ex:1: FishjamWeb.PeerController.phoenix_controller_pipeline/2
       (phoenix 1.7.12) lib/phoenix/router.ex:484: Phoenix.Router.__call__/5
       (fishjam 0.6.3) lib/fishjam_web/endpoint.ex:1: FishjamWeb.Endpoint.plug_builder_call/2
       (fishjam 0.6.3) lib/fishjam_web/endpoint.ex:1: FishjamWeb.Endpoint.call/2
       (phoenix 1.7.12) lib/phoenix/test/conn_test.ex:225: Phoenix.ConnTest.dispatch/5
       test/fishjam_web/controllers/peer_controller_test.exs:134: (test)

     The following output was logged:
     12:53:12.978 room_id=1e26932f-8f32-4cf2-9159-f48e120b82ef [info] Initialize room
     12:53:12.978 [info] Created room "1e26932f-8f32-4cf2-9159-f48e120b82ef"
     12:53:12.978 room_id=1e26932f-8f32-4cf2-9159-f48e120b82ef [info] Added peer "9689ddf6-fbb1-43c7-afa5-8636023fceac"
     12:53:12.997 room_id=1e26932f-8f32-4cf2-9159-f48e120b82ef [info] Removed peer "9689ddf6-fbb1-43c7-afa5-8636023fceac" from room "1e26932f-8f32-4cf2-9159-f48e120b82ef"
     12:53:12.997 [info] Deleted room "1e26932f-8f32-4cf2-9159-f48e120b82ef"


```

lib/fishjam/rpc_client.ex
```diff
@@ -48,6 +48,7 @@ defmodule Fishjam.RPCClient do
 
         acc
     end)
+    |> List.flatten()
```
Rados13, line 51: Why do we need that?

test/fishjam/cluster/api_test.exs
Rados13: I agree also we could think about more helper functions, which would wrap parts of logic and these new tests could be simply calling this one function.

test/fishjam/cluster/api_test.exs
```diff
@@ -0,0 +1,248 @@
...
+
+    room_id = response_body1 |> Jason.decode!() |> get_in(["data", "room", "id"])
+
+    delete_room(fishjam_instance1, room_id)
+
+    assert_room_count_on_fishjam(fishjam_instance1, 0)
+    assert_room_count_on_fishjam(fishjam_instance2, 1)
+
+    assert_room_count_in_cluster(node1, 1)
+    assert_room_count_in_cluster(node2, 1)
```
Rados13, line 67: I think this could be moved to a helper function which would be provided with a number of rooms on each node and it could verify that each node has a correct number of rooms and whole cluster API also.

test/fishjam/cluster/api_test.exs
```diff
@@ -0,0 +1,248 @@
...
+    assert_room_count_in_cluster(other_node, 1)
+
+    room_id = response_body1 |> Jason.decode!() |> get_in(["data", "room", "id"])
+
+    _response_body2 = add_hls_component(other_node, room_id)
+
+    _response_body3 = add_file_component(other_node, room_id)
+
+    # Wait a while for segments and manifest to get created
+    Process.sleep(20_000)
```
Rados13, line 131: image

test/fishjam_web/controllers/peer_controller_test.exs
```diff
@@ -80,9 +82,9 @@ defmodule FishjamWeb.PeerControllerTest do
     end
 
     test "renders errors when room doesn't exist", %{conn: conn} do
-      room_id = "invalid_room"
+      room_id = ID.generate()
       conn = post(conn, ~p"/room/#{room_id}/peer", type: @peer_type)
-      assert json_response(conn, :not_found)["errors"] == "Room #{room_id} does not exist"
+      assert json_response(conn, :not_found)["errors"] == "Room with this ID does not exist"
```
Rados13, line 87: I think it would be beneficial that we would return the passed ID, so the client could verify if it is a typo or what.

test/fishjam_web/controllers/component_controller_test.exs
```diff
@@ -73,11 +75,11 @@ defmodule FishjamWeb.ComponentControllerTest do
     end
 
     test "deletes component from not exisiting room", %{conn: conn, component_id: component_id} do
-      room_id = "abc"
+      room_id = ID.generate()
```
Rados13, line 78: Why do we need to make this change?

config/test.exs
```diff
@@ -5,7 +5,12 @@ config :fishjam,
   port: 4002,
   server_api_token: "development",
   webrtc_metrics_scrape_interval: 50,
-  room_metrics_scrape_interval: 1
+  room_metrics_scrape_interval: 1,
+  feature_flags: [
+    # TODO: Enable this flag here once we start using it in production
+    custom_room_name_disabled: false
+  ],
```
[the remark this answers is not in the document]
sgfn: Nah, that's an oversight -- it should be `request_routing_enabled?` here

lib/fishjam/room/id.ex
```diff
@@ -8,20 +8,17 @@ defmodule Fishjam.Room.ID do
...
   @spec determine_node(id()) ::
-          {:ok, node()} | {:error, :invalid_room_id} | {:error, :invalid_node}
+          {:ok, node()} | {:error, :invalid_room_id | :node_not_found}
   def determine_node(room_id) do
     with {:ok, room_id} <- validate_room_id(room_id),
-         node_name <- decode_node_name(room_id),
+         {:ok, node_name} <- decode_node_name(room_id),
          true <- node_present_in_cluster?(node_name) do
-      {:ok, node_name}
+      {:ok, String.to_existing_atom(node_name)}
```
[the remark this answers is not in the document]
sgfn: Consider the following scenario:
1. I make a `GET /room/a-b-c-d-666f6f626172` request to the Fishjam
2. `determine_node` sees that the room ID is valid, so it calls `decode_node_name`
3. `decode_node_name` decodes the node name part to `"foobar"`, tries to create an existing atom and crashes the app

That's precisely what I wanted to avoid with this approach

lib/fishjam/rpc_client.ex
sgfn: When listing rooms, every node returns a list. Without this call, the result was a nested list of lists.

test/fishjam/cluster/api_test.exs
sgfn: Any other ideas are welcome XD

test/fishjam_web/controllers/component_controller_test.exs
sgfn: Without it, tests like this one failed on the `decode_node_name` part in `room/id.ex` (when running with `REQUEST_ROUTING=true`)

test/fishjam_web/controllers/peer_controller_test.exs
sgfn: Unfortunately, it's a PITA to achieve with the fallback controller...

lib/fishjam/cluster/room_service.ex
```diff
@@ -0,0 +1,78 @@
...
+
+      nodes_resources ->
+        min_node = find_best_node(nodes_resources)
+        Logger.info("Node with least used resources is #{inspect(min_node)}")
+
+        case RPCClient.call(min_node, @local_module, :create_room, [config]) do
+          {:ok, result} -> result
+          :error -> {:error, :rpc_failed}
+        end
+    end
```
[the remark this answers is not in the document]
sgfn: Refactored

lib/fishjam/rpc_client.ex
Rados13: Okay, but this is valid only in the case of listing a room; in other cases, it doesn't have to, so I think this is not a good place to use it.

test/fishjam_web/controllers/component_controller_test.exs
Rados13: Ok, so from this PR, we remove the possibility of passing custom room ID?

test/fishjam/cluster/api_test.exs
Rados13: I think we could subscribe on server notifications and await on `HlsPlayable`; this way, it should be a little bit shorter.

lib/fishjam_web/controllers/component_controller.ex
kamciokodzi: Ah it works if there is no `else` clause completely then results goes to fallback, my bad :)

test/fishjam/cluster/api_test.exs
```diff
@@ -139,60 +124,54 @@ defmodule Fishjam.Cluster.ApiTest do
     assert {:ok, %HTTPoison.Response{status_code: 201, body: body}} =
              HTTPoison.post("http://#{fishjam_instance}/room", request_body, @post_headers)
 
-    body
+    room_data = body |> Jason.decode!() |> Map.fetch!("data")
+    room_node = get_fishjam_address(room_data)
+
+    {room_data, room_node}
```
Rados13, line 130: Could, instead of a tuple, return a map with fields like `data,` `node,` `id` etc, and then pattern match in a specific line of code.

config/test.exs
```diff
@@ -5,7 +5,12 @@ config :fishjam,
   port: 4002,
   server_api_token: "development",
   webrtc_metrics_scrape_interval: 50,
-  room_metrics_scrape_interval: 1
+  room_metrics_scrape_interval: 1,
+  feature_flags: [
+    # TODO: Enable this flag here once we start using it in production
+    request_routing_enabled?: false
```
kamciokodzi, line 11: I feel like this is unecessary since we have this in `runtime.exs` and also default value is set there.

lib/fishjam/cluster/room_service.ex
```diff
@@ -0,0 +1,72 @@
...
+  alias Fishjam.{FeatureFlags, Room, RPCClient}
+
+  @local_module Fishjam.Local.RoomService
+
+  @impl true
+  def create_room(config) do
+    node_resources = RPCClient.multicall(@local_module, :get_resource_usage)
+    min_node = find_best_node(node_resources)
+
+    if node_resources == [],
```
kamciokodzi, line 19: I wouldnt write if for case like that, this shouldn't ever happen -> call to self should always succeed so there will be at least 1 element on list right?

sgfn: I agree but @Rados13 wanted to handle this case

lib/fishjam/component/hls/cluster/request_handler.ex
```diff
@@ -0,0 +1,39 @@
...
+  def handle_manifest_request(room_id, partial),
+    do: route_request(room_id, :handle_manifest_request, [room_id, partial])
+
+  @impl true
+  def handle_delta_manifest_request(room_id, partial),
+    do: route_request(room_id, :handle_delta_manifest_request, [room_id, partial])
+
+  defp route_request(room_id, fun, args) do
+    with {:ok, node} <- Room.ID.determine_node(room_id),
+         {:here?, false} <- {:here?, node == Node.self()},
```
kamciokodzi, line 30: @Rados13 had a nice idea about this one. Instead of checking if node is self and then apply rpc or local mfa -> always use RPC even to self, so this one path in with will disappear.

sgfn: An awesome idea; in this case, though, we call different functions depending on whether the HLS content is on this node or another one -- so it can't really be done

kamciokodzi: Yeah everywhere except `hls` I think this rule should apply :D

test/fishjam_web/controllers/peer_controller_test.exs
sgfn: The room ID was extracted from params passed to the regular controller, and the fallback controller received a 3-tuple with the error string, where the room ID was already baked in. I don't know if it's possible to do that in the fallback controller without restoring the `else` cases that were removed for brevity and eliminating code duplication

test/fishjam_web/controllers/component_controller_test.exs
sgfn: No, why? The possibility is still there, this just had to do with the fact that this test could be run both with and without request routing, and so I fixed it to pass in both cases

test/fishjam/cluster/api_test.exs
```diff
@@ -34,103 +34,88 @@ defmodule Fishjam.Cluster.ApiTest do
...
   @tag timeout: @max_test_duration
-  test "load-balancing and listing rooms in a cluster", %{nodes: [node1, node2]} do
-    response_body1 = add_room(node1)
+  test "adding a single room in a cluster", %{nodes: [node1, node2]} do
+    {_room_data, room_node} = add_room(node1)
 
-    fishjam_instance1 = get_fishjam_address(response_body1)
-
-    assert_room_count_on_fishjam(fishjam_instance1, 1)
+    other_node = if room_node == node1, do: node2, else: node1
```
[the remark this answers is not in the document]
sgfn: I'm not sure I get you. What exactly would you like to achieve?

lib/fishjam/component/hls/cluster/request_handler.ex
sgfn: Well then that's already implemented...

test/fishjam_web/controllers/peer_controller_test.exs
sgfn: Refactored

test/fishjam/room/id_test.exs
```diff
@@ -22,7 +22,10 @@ defmodule Fishjam.Room.IDTest do
     test "returns error if node is not detected in cluster" do
       invalid_node = :invalid_node |> Atom.to_string() |> Base.encode16(case: :lower)
       invalid_room_id = "room-id-#{invalid_node}"
-      assert {:error, :node_not_found} == Subject.determine_node(invalid_room_id)
+
+      if Fishjam.FeatureFlags.request_routing_enabled?() do
```
Rados13, line 26: I think we should test both cases, but I will introduce it in my PR #211 .

Rados13 (approved):

PR #213, opened 2024-06-13
feat: cluster health report

kamciokodzi:

Updated Elixir SDK: https://github.com/fishjam-dev/elixir_server_sdk/pull/70
Updated Python SDK: https://github.com/fishjam-dev/python-server-sdk/pull/43


lib/fishjam/rpc_client.ex
```diff
@@ -30,9 +37,16 @@ defmodule Fishjam.RPCClient do
...
+    start_time = System.monotonic_time(:millisecond)
+
     nodes()
     |> :erpc.multicall(module, function, args, timeout)
     |> handle_result()
+    |> then(fn result ->
+      emit_rpc_duration_event([:fishjam, :rcp_client, :multicall], start_time)
+
+      result
+    end)
```
Rados13, line 49: I think we could use [tap function](https://hexdocs.pm/elixir/1.17.0/Kernel.html#tap/2) instead.
```suggestion
    |> tap(&emit_rpc_duration_event([:fishjam, :rcp_client, :multicall], start_time))
```
```suggestion
    |> then(fn result ->
      emit_rpc_duration_event([:fishjam, :rcp_client, :multicall], start_time)

      result
    end)
```

lib/fishjam/rpc_client.ex
```diff
@@ -51,4 +65,9 @@ defmodule Fishjam.RPCClient do
   end
 
   defp nodes, do: [Node.self() | Node.list()]
+
+  defp emit_rpc_duration_event(event_name, start_time) do
+    duration = System.monotonic_time(:millisecond) - start_time
```
Rados13, line 70: What do you think about adding a function `get_time`, so we will guarantee the time units to be the same in this module?

lib/fishjam_web/telemetry.ex
```diff
@@ -108,6 +108,25 @@ defmodule FishjamWeb.Telemetry do
...
+
+        counter("fishjam.rpc_client.call.count",
+          event_name: [:fishjam, :rpc_client, :call],
+          measurement: :duration
+        ),
+        counter("fishjam.rpc_client.multicall.count",
+          event_name: [:fishjam, :rpc_client, :multicall],
+          measurement: :duration
+        ),
+        metric_type.("fishjam.rpc_client.call.duration.milliseconds",
```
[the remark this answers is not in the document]
sgfn: Yeah, that's pretty much what has to be done, unfortunately.

`last_value` isn't really very useful to us, it gives us only one reading at a time (the duration of the last recorded request/RPC in this case), which, since we scrape the metrics from the Prometheus endpoint every 10 seconds, isn't really representative. [`summary` performs some aggregations](https://prometheus.io/docs/practices/histograms/) and is better suited here, but the library doesn't support it

lib/fishjam_web/telemetry.ex
```diff
@@ -108,6 +108,25 @@ defmodule FishjamWeb.Telemetry do
         #   - WebSocket traffic
         #   - RTP events (RTSP components don't use ICE)
         #   - HTTP traffic related to metrics (not handled by Phoenix)
+
+        counter("fishjam.rpc_client.call.count",
+          event_name: [:fishjam, :rpc_client, :call],
+          measurement: :duration
+        ),
+        counter("fishjam.rpc_client.multicall.count",
```
sgfn, line 116: These metrics should be suffixed `_total`, according to the [Prometheus style guide](https://prometheus.io/docs/practices/naming/)

Additionally, I don't think we need to specify `rpc_client` in the metric name -- it's the internal name of one of our modules, and the `client` part doesn't really communicate anything of value. How do you feel about `fishjam_rpc_call_total` and `fishjam_rpc_multicall_total` instead?

lib/fishjam_web/telemetry.ex
```diff
@@ -108,6 +108,25 @@ defmodule FishjamWeb.Telemetry do
...
+        ),
+        metric_type.("fishjam.rpc_client.call.duration.milliseconds",
+          event_name: [:fishjam, :rpc_client, :call],
+          measurement: :duration,
+          unit: :millisecond
+        ),
+        metric_type.("fishjam.rpc_client.multicall.duration.milliseconds",
+          event_name: [:fishjam, :rpc_client, :multicall],
+          measurement: :duration,
+          unit: :millisecond
```
sgfn, line 128: The style guide also recommends always using base units, so in this case `seconds`

lib/fishjam_web/controllers/healthcheck_controller.ex
```diff
@@ -23,14 +23,23 @@ defmodule FishjamWeb.HealthcheckController do
...
     |> put_resp_content_type("application/json")
-    |> render("show.json", report: report)
+    |> send_resp(200, Jason.encode!(%{data: report}))
   end
 
   defp get_health_report() do
     %{
-      status: :up,
+      localStatus: get_health_status(),
+      distributionEnabled: Application.fetch_env!(:fishjam, :dist_config)[:enabled],
```
Rados13, line 32: Shouldn't we add here `Map.fetch!` to prevent a situation in which this returns nil?

lib/fishjam/rpc_client.ex
```diff
@@ -51,4 +65,9 @@ defmodule Fishjam.RPCClient do
...
+    :telemetry.execute(event_name, %{duration: duration})
```
sgfn, line 71: Again, this should be in seconds; you can use `System.monotonic_time()` and perform the conversion automatically by specifying `unit: {:native, :second}` in `telemetry.ex`, just like in the Phoenix metrics

lib/fishjam_web/api_spec/health_report.ex
```diff
@@ -48,12 +36,18 @@ defmodule FishjamWeb.ApiSpec.HealthReport do
...
+      localStatus: NodeStatus,
+      nodesInCluster: %Schema{type: :integer, description: "Number of nodes in cluster"},
+      distributionEnabled: %Schema{
+        type: :boolean,
+        description: "Cluster distribution enabled/disabled"
+      },
+      nodesStatus: %Schema{
+        type: :array,
+        items: NodeStatus,
+        description: "Status of each node in cluster"
```
Rados13, line 48: Shouldn't we mention here that this node is not included?

lib/fishjam/rpc_client.ex
```diff
@@ -15,11 +15,18 @@ defmodule Fishjam.RPCClient do
...
     try do
-      {:ok, :erpc.call(node, module, function, args, timeout)}
+      result = :erpc.call(node, module, function, args, timeout)
+      emit_rpc_duration_event([:fishjam, :rcp_client, :call], start_time)
+
+      {:ok, result}
     rescue
       e ->
         Logger.warning("RPC call to node #{node} failed with exception: #{inspect(e)}")
+        emit_rpc_duration_event([:fishjam, :rcp_client, :call], start_time)
```
sgfn, line 28: typo: `rcp_client` 😄

lib/fishjam_web/telemetry.ex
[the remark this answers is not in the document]
Rados13: I'm not sure, but based [on this line](https://github.com/fishjam-dev/fishjam/blob/b264fed329ffc9d233a07df60a046ec6f6c86c95/lib/fishjam_web/telemetry.ex#L27) I think that they are presented with `metric_type` `last_value`.

lib/fishjam/rpc_client.ex
sgfn, line 49: Why not simply
```elixir
result =
  nodes()
  |> ...

emit_rpc_duration_event(...)
result
```
?

lib/fishjam_web/controllers/healthcheck_controller.ex
```diff
@@ -23,14 +23,23 @@ defmodule FishjamWeb.HealthcheckController do
...
+      nodesInCluster: length([Node.self() | Node.list()]),
+      nodesStatus: Fishjam.RPCClient.multicall(__MODULE__, :get_health_status, [])
```
sgfn, line 34: This line is problematic IMO. A healthcheck endpoint is expected to respond quickly (for instance, we use it to assess the liveness of Fishjam running in a container), here the multicall will slow it down significantly.

I'd suggest adding a separate endpoint for a cluster-wide healthcheck like this instead of extending this one.

sgfn (changes requested):
Please make a PR in the docs as well (see [this page](https://fishjam-dev.github.io/fishjam-docs/getting_started/metrics))

Something to consider: wouldn't we like an additional metric keeping track of all the failed RPC calls?

lib/fishjam_web/api_spec/health_report.ex
kamciokodzi: This node is included.

lib/fishjam_web/controllers/healthcheck_controller.ex
kamciokodzi: As long as the instances are in the same DC the time to make RPC is `negligible`.

lib/fishjam_web/controllers/healthcheck_controller.ex
kamciokodzi: Well there is no need for that based on what is in config reader:
`lib/fishjam/config_reader.ex:208`

